PR #7, opened 2022-05-12
Use pull_request_target for e2e tests.

wlynch:
#### Summary

pull_request doesn't allow for usage of OIDC tokens, so the test
sign+verify workflow can't run. Although this is generally discouraged
by GitHub (see
https://securitylab.github.com/research/github-actions-preventing-pwn-requests/),
this is useful as an e2e testing tool for testing.

#### Ticket Link

n/a

#### Release Note

```release-note
None
```

### DIFF
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -1,0 +1,70 @@
+name: E2E
+
+on:
+  push:
+    branches: ['main']
+  pull_request_target:
+    branches: ['main']
+  workflow_dispatch:
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    permissions:
+      id-token: write  # Enable OIDC
+
+      # The rest of these are sanity-check settings, since I'm not sure if the
+      # org default is permissive or restricted. 
+      # See https://docs.github.com/en/actions/security-guides/automatic-token-authentication#permissions-for-the-github_token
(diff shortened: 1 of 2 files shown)


.github/workflows/e2e.yaml
```diff
@@ -0,0 +1,68 @@
+name: E2E
+
+on:
+  push:
+  pull_request_target:
```
imjasonh, line 5:
```suggestion
on:
  push:
    branches: ['main']
  pull_request_target:
    branches: ['main']
```

nsmith5 (approved):
Noice

.github/workflows/go.yml
```diff
@@ -2,17 +2,12 @@ name: Go
 
 on:
   push:
-    branches: [ main ]
```
imjasonh, line 5: I'd undo this, so we still only run these tests on pushes to main and PRs to main. Otherwise you get CI running in forks when folks push to a branch in their fork.

.github/workflows/e2e.yaml
wlynch: Done.

.github/workflows/go.yml
wlynch: Done.

imjasonh (approved):

PR #14, opened 2022-05-12
Add GITSIGN_LOG environment variable

wlynch:
#### Summary

Add GITSIGN_LOG environment variable for debug log path.

Makes debug log behavior opt-in, and lets users control where the logs
are written.

#### Ticket Link

n/a

#### Release Note

```release-note
Environment variable GITSIGN_LOG can be set to log status output to file for debugging.
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,13 +67,29 @@ Date:   Mon May 2 16:51:44 2022 -0400
 
 ## Limitations
 
-- Due to how Git invokes signing tools, `gitsign` does not work in device
-  mode - a browser session is required to sign commits.
+- When Git invokes signing tools, both stdout and stderr are captured which
+  means `gitsign` cannot push back messages to shells interactively. Because of
+  this, device mode does not work with `gitsign` - a browser capable session is
+  required to sign commits.
+
+## Debugging
+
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -67,13 +67,29 @@ Date:   Mon May 2 16:51:44 2022 -0400
...
+## Debugging
+
+If there is an error during signing, you may see an error like:
+
+```
+error: gpg failed to sign the data
+fatal: failed to write commit object
+```
+
+Because of [`Limitations`](#limitations) with Git signing tools, since `gitsign`
```
imjasonh, line 84: Is it possible that sensitive contents could be logged? Could we make this opt-in, and opt in during our tests?

wlynch: I don't think we do, but there's not enforcement on that. How does an env variable for the path sound?

nsmith5: I like opt-in. Don't write things anywhere by default and e.g `--log oh-dear-what-has-gone-wrong.log` when things go bad. Or `GITSIGN_LOG=how-did-we-git-here.log`

wlynch: Done!

nsmith5 (approved):

PR #25, opened 2022-05-14
add initial makefile

### LINKED ISSUE #24: `gitsign --version` shows `unknown`
Hi,

I just started to play a bit with `gitsign` to start packaging it for `nixpkgs` and I noticed that when using the binaries in the release assets the version was not set:
```
$ ./gitsign_0.0.2-alpha_linux_amd64 --version
unknown
```

I did not check the build pipeline but since it relies on VCS information retrieved from the `debug/buildinfo` I am guessing it will not play well with 107ac2485733ef71c8dce26ade70c68879bcdb45 anyway.

https://github.com/sigstore/gitsign/blob/107ac2485733ef71c8dce26ade70c68879bcdb45/main.go#L76-L88

cpanato:
#### Summary
add initial makefile to build and install `gitsign` 

using the `make install` you will be able to have the command `gitsign --version` to return the git hash if use `go install github.com/sigstore/gitsign@latest` this does not populate it

#### Ticket Link

Fixes #24 

#### Release Note

```release-note
NONE
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,24 @@
+#
+# Copyright 2022 The Sigstore Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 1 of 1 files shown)


wlynch (approved):
Thanks! :D

PR #39, opened 2022-05-19
Added environment variable for OIDC Redirect URL

Boolman:
Allow for specifying the OIDC Redirect URL so we can specify a predictable redirect uri, this is needed for Keycloak for example

### DIFF
--- a/internal/fulcio/fulcio.go
+++ b/internal/fulcio/fulcio.go
@@ -48,10 +48,11 @@ func NewIdentity(ctx context.Context, w io.Writer) (*Identity, error) {
 		authFlow = fulcio.FlowToken
 	}
 	sv, err := sign.SignerFromKeyOpts(ctx, "", "", options.KeyOpts{
-		FulcioURL:    envOrValue("GITSIGN_FULCIO_URL", "https://fulcio.sigstore.dev"),
-		OIDCIssuer:   envOrValue("GITSIGN_OIDC_ISSUER", "https://oauth2.sigstore.dev/auth"),
-		OIDCClientID: clientID,
-		RekorURL:     envOrValue("GITSIGN_REKOR_URL", "https://rekor.sigstore.dev"),
+		FulcioURL:       envOrValue("GITSIGN_FULCIO_URL", "https://fulcio.sigstore.dev"),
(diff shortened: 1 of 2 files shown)


cpanato (dismissed):

internal/fulcio/fulcio.go
```diff
@@ -48,10 +48,11 @@ func NewIdentity(ctx context.Context, w io.Writer) (*Identity, error) {
...
 	}
 	sv, err := sign.SignerFromKeyOpts(ctx, "", "", options.KeyOpts{
-		FulcioURL:    envOrValue("GITSIGN_FULCIO_URL", "https://fulcio.sigstore.dev"),
-		OIDCIssuer:   envOrValue("GITSIGN_OIDC_ISSUER", "https://oauth2.sigstore.dev/auth"),
-		OIDCClientID: clientID,
-		RekorURL:     envOrValue("GITSIGN_REKOR_URL", "https://rekor.sigstore.dev"),
+		FulcioURL:       envOrValue("GITSIGN_FULCIO_URL", "https://fulcio.sigstore.dev"),
+		OIDCIssuer:      envOrValue("GITSIGN_OIDC_ISSUER", "https://oauth2.sigstore.dev/auth"),
+		OIDCClientID:    clientID,
+		OIDCRedirectURL: envOrValue("GITSIGN_OIDC_REDIRECT_URL", "http://localhost:5000/auth/callback"),
```
wlynch, line 54:
```suggestion
		OIDCRedirectURL: os.Getenv("GITSIGN_OIDC_REDIRECT_URL"),
```

Redirect URL is a bit unique in that there's already code for handling defaults - which is also smart about auto-selecting an unused port. The other environment variables are defaults that are set by cosign (though @imjasonh perhaps we should also move these up to sigstore/sigstore 🤔 )

https://github.com/sigstore/sigstore/blob/5095b189ff654aea7e4964692a082855714e83b5/pkg/oauth/oidc/interactive.go#L56

wlynch (changes requested):
Thanks for adding this! One small tweak, but otherwise looks good.

wlynch (approved):

PR #56, opened 2022-06-02
Add load bearing space to context variable.

wlynch:
#### Summary

Not 100% sure if this is correct, but latest tests are not properly
evaluating this. GitHub docs say this should be `${{ <context> }}` so
trying to see if fixing the space fixes it.

#### Release Note

```release-note

```


wlynch:
nevermind! this does work as is, but not for pushes since there is no PR number there. Will think this through some more.

PR #58, opened 2022-06-02
Remove dependency on cosign/cli/rekor.

wlynch:
#### Summary

Keeps the dependency on pkg/cosign, since there's a lot of useful
behavior for interacting with the TLog (we probably want to pull this
out into its own sdk somewhere else).

#### Release Note

```release-note
NONE
```


imjasonh (approved):

wlynch:
not needed anymore because of https://github.com/sigstore/gitsign/commit/c72bc2cc8d5cae883603125a837934d16072b179#diff-b45aace94d5abd146576cc645831a51a5f0772e385afe168433868fdd37d2dd1

PR #60, opened 2022-06-02
Export rekor package.

wlynch:
#### Summary

This is used within pkg/git, so exporting the matching interface to stay
consistent.

Did my best to make this fairly generic. @imjasonh I know you have opinions about how much we export here, so would love to get your thoughts here. I'd like for other clients to be able to verify commits produced by gitsign, and rekor verification is part of that. I wonder how much of this should go in here vs `sigstore/sigstore` though 🤔 

#### Release Note

```release-note
Exported Go client for Rekor validation of commits.
```

### DIFF
--- a/pkg/rekor/rekor.go
+++ b/pkg/rekor/rekor.go
@@ -49,7 +49,7 @@ type Verifier interface {
 
 // Writer represents a mechanism to write content to Rekor.
 type Writer interface {
-	Write(ctx context.Context, sig []byte, commitSHA string, cert *x509.Certificate) (*models.LogEntryAnon, error)
+	Write(ctx context.Context, commitSHA string, sig []byte, cert *x509.Certificate) (*models.LogEntryAnon, error)
 }
 
 // Client implements a basic rekor implementation for writing and verifying Rekor data.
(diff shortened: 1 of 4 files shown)


nsmith5 (approved):

pkg/rekor/rekor.go
```diff
@@ -43,21 +42,25 @@ import (
 	"github.com/sigstore/sigstore/pkg/cryptoutils"
 )
 
+// Verifier represents a mechanism to get and verify Rekor entries.
 type Verifier interface {
-	Get(ctx context.Context, commit string, cert *x509.Certificate) (*models.LogEntryAnon, error)
+	Get(ctx context.Context, data []byte, cert *x509.Certificate) (*models.LogEntryAnon, error)
```
imjasonh, line 47: This `rekor.Verifier` interface doesn't seem very specific to gitsign. If it was verifying commit SHAs specifically that'd make sense. Otherwise, maybe this belongs in sigstore/sigstore (or sigstore/rekor) so others can use it.

OTOH, I'm a big fan of not exporting something, in here or in broader sigstore, until there's a client, so you know what you need from it. If you have a client and this is the interface that makes sense there, this sounds fine to me.

gitsign's Go package API compatibility promises seem quite a bit more lax than sigstore/sigstore's, so if we need to change this later (or unexport it) I think we can.

wlynch: wdyt about slimming this down to:

```go
type Verifier interface {
	Verify(ctx context.Context, commitSHA string, cert *x509.Certificate) (*models.LogEntryAnon, error)
}
```
?

The main reason why this was split was so that we could tell the difference between "there was no tlog entry" and "we couldn't validate the tlog entry we found". But we could also surface this type of information within the error itself.

wlynch: Went ahead and made this change. lmk what you think!

imjasonh (approved):
Sorry for not getting back sooner, looks great to me! 👍

PR #67, opened 2022-06-10
Use pkg/fulcioroots from sigstore/sigstore

imjasonh:
Signed-off-by: Jason Hall <jason@chainguard.dev>

These methods were moved from cosign->sigstore in https://github.com/sigstore/sigstore/pull/435/

After this change, remaining dependencies on cosign are:

```
$ git grep sigstore/cosign
...
internal/fulcio/identity.go:    "github.com/sigstore/cosign/pkg/providers"
internal/rekor/rekor.go:        "github.com/sigstore/cosign/pkg/cosign"
main.go:        _ "github.com/sigstore/cosign/pkg/providers/all"
```

#### Release Note

```release-note
NONE
```


wlynch (changes requested):
LGTM! Just need to fix up the test.

imjasonh:
> LGTM! Just need to fix up the test.

Hmm, I didn't realize this test was using the fake roots code from cosign. We had decided to keep that in cosign for now, since we thought nobody but cosign would need it, but yet here we are 🙃.

We'll either need to find a way to get this test not to need this fake CA setup, or to move the fallback logic into the version in sigstore/sigstore. WDYT?

wlynch:
SGTM! 

Leaning towards making this change within gitsign instead of sigstore/sigstore - I'm thinking it might be valuable to have the VerifySignature have configurable roots (either as a variadic option or moving func to a struct that has the roots / clients configured) instead of depending on global state from fulcioroots. 🤔

wlynch (approved):

PR #75, opened 2022-06-16
Implement Credential Caching

### LINKED ISSUE #21: Batch signing support
**Description**

This is probably a broader sigstore feature request, but creating an issue here to track.

An annoyance with using gitsign today is you need to hop through the oauth flow for every commit - this can be very tedious for quick `git commit --amend`s or multi-commit rebases.

We'll want to look into some way to cache the signing keypair (or possibly the Fulcio OIDC token) to support multiple operations for the lifetime of the keypair (~10 mins) without needing to hop through the flow again and get a new key. This key shouldn't hit disk.

https://git-scm.com/docs/git-credential-cache strikes me as something similar to model off of for ideas.

Related discussion: https://sigstore.slack.com/archives/C01PZKDL4DP/p1652208232815029

wlynch:
#### Summary

This introduces a socket based credential cache to allow caching of keys
so that users do not need to go through the OIDC flow multiple times
for batch operations (e.g. rebases).

Credentials are keyed to git working directories so that different repos can
cache different identities, and credentials are never directly stored to disk.

#### Ticket Link

Fixes #21 

#### Release Note

```release-note
Adds gitsign-credential-cache: an optional credential cache for reusing keys for multiple signing requests.
```


cmd/gitsign-credential-cache/README.md
```diff
@@ -0,0 +1,25 @@
+# gitsign-credential-cache
+
+`gitsign-credential-cache` is a helper binary that allows users to cache signing
+credentials. This can be helpful in situations where you need to perform
+multiple signing operations back to back.
+
+Credentials are stored in memory, and the cache is exposed via a Unix socket.
+Credentials stored in this cache are only as secure as the unix socket
+implementation on your OS - any user that can access the socket can access the
+data. When in doubt, we recommend **not** using the cache.
```
imjasonh, line 10: Just to be clear, it's disabled by default right? Can we make that more explicit?

cmd/gitsign-credential-cache/main.go
```diff
@@ -0,0 +1,43 @@
...
+	"os"
+	"path/filepath"
+
+	"github.com/sigstore/gitsign/internal/cache"
+	cachepb "github.com/sigstore/gitsign/internal/cache/cache_go_proto"
+	"google.golang.org/grpc"
+)
+
+func main() {
+	home, err := os.UserHomeDir()
```
imjasonh, line 16: (not blocking) Why do I see people using https://github.com/mitchellh/go-homedir when this is in the stdlib? Is go-homedir better in some way? 🤔

go.mod
```diff
@@ -10,13 +10,16 @@ require (
...
 	github.com/google/go-cmp v0.5.8
+	github.com/patrickmn/go-cache v2.1.0+incompatible
 	github.com/pborman/getopt v0.0.0-20180811024354-2b5b3bfb099b
 	github.com/sigstore/cosign v1.8.1-0.20220601172726-ae90c7495df6
 	github.com/sigstore/fulcio v0.1.2-0.20220114150912-86a2036f9bc7
 	github.com/sigstore/rekor v0.8.0
 	github.com/sigstore/sigstore v1.2.1-0.20220614141825-9c0e2e247545
 	golang.org/x/crypto v0.0.0-20220411220226-7b82a4e95df4
 	golang.org/x/oauth2 v0.0.0-20220524215830-622c5d57e401
+	google.golang.org/grpc v1.47.0
```
imjasonh, line 21: Any big reason to prefer gRPC+proto over HTTP+JSON here? You're basically just POSTing some bytes and GETting some bytes, proto seems like a large hammer for that. Unless you have broader plans for this in the future?

internal/fulcio/identity.go
```diff
@@ -26,25 +26,64 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"path/filepath"
 
+	"github.com/sigstore/cosign/cmd/cosign/cli/fulcio/fulcioroots"
```
imjasonh, line 31:
```suggestion
	"github.com/sigstore/sigstore/pkg/fulcioroots"
```

This will also mean `fulcioroots.Get()` returns an err.

cmd/gitsign-credential-cache/main.go
wlynch: os.UserHomeDir() was added in Go 1.12 (~May 2019). Where as go-homedir was created in 2015.

go.mod
wlynch: Moved to net/rpc for now - HTTP seemed overkill for something that would only ever be local (grpc at least codegens away much of the complexity of setting up the client/server).

internal/fulcio/identity.go
wlynch: Done!

cmd/gitsign-credential-cache/README.md
wlynch: Took another pass at this.

cmd/gitsign-credential-cache/README.md
```diff
@@ -0,0 +1,36 @@
...
+implementation on your OS - any user that can access the socket can access the
+data.
+
+⚠️ When in doubt, we recommend **not** using the cache. In particular:
+
+- If you're running on a shared system
+  - if other admins have access to the cache socket they can access your keys.
+- If you're running in an environment that has ambient OIDC credentials (e.g.
+  GCE/GKE, AWS, GitHub Actions, etc.) - Gitsign will automatically use the
+  environment's OIDC credentials. You don't need caching.
```
imjasonh, line 18:
```suggestion
- If you're running in an environment that has ambient OIDC credentials (e.g.
  GCE/GKE, AWS, GitHub Actions, etc.)
  - Gitsign will automatically use the
  environment's OIDC credentials. You don't need caching.
```

cmd/gitsign-credential-cache/README.md
```diff
@@ -0,0 +1,36 @@
...
+
+If you understand the risks, read on!
+
+## What's stored in the cache
+
+- Ephemeral Private Key
+- Fulcio Code Signing certificate + chain
+
+Data is stored keyed to your Git working directory (i.e. different repo paths
+will cache different keys)
```
imjasonh, line 28:
```suggestion
## What's stored in the cache

- Ephemeral Private Key
- Fulcio Code Signing certificate + chain

All data is stored in memory, keyed to your Git working directory (i.e. different repo paths
will cache different keys)

The data that is cached would allow any user with access to sign artifacts as you, until the signing certificate expires, typically in ten minutes.
```

imjasonh (dismissed):

znewman01 (changes requested):
LGTM! Only blocking comment is about permissions of the socket

cmd/gitsign-credential-cache/main.go
```diff
@@ -0,0 +1,56 @@
...
+	if err != nil {
+		log.Fatalf("error getting user home directory: %v", err)
+	}
+
+	dir := filepath.Join(home, ".sigstore", "gitsign")
+	if err := os.MkdirAll(dir, 0700); err != nil {
```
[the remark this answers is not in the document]
znewman01: Can you convince me the socket file permissions are appropriately restricted when we create the socket as below? I tried to find out what `Listen` does for permissions but I couldn't find it anywhere.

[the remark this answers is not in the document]
wlynch: Good catch - set the default umask to restrict permissions on all files to guarantee that we only create files/directories scoped to the current user.

imjasonh (approved):
Let's do it!

When this goes in I'll try it out for a few days and see if anything breaks. 🤞

PR #92, opened 2022-07-05
--version: Print out relevant env variables.

wlynch:
#### Summary

As an extra debugging tool, print out environment variables with known
prefixes that might affect tool behavior as part of --version.

#### Release Note

```
--version now prints out environment variables with known prefixes that might affect tool behavior
```

#### Documentation

Done.


imjasonh (approved):

wlynch:
> Should we have cosign do this too?

Probably! It's a low-effort way of surfacing config information that might be causing unexpected behavior. I can make a PR.

PR #102, opened 2022-08-01
Add note to credential cache docs about cache directory selection.

wlynch:
#### Summary

Also fixes gitignore to not exclude the credential cache directory
itself.

#### Release Note

NONE

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ dist/*
 .vscode/*
 gitsign
 gitsign-credential-cache
+!/cmd/gitsign-credential-cache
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -4,3 +4,4 @@ dist/*
 .vscode/*
 gitsign
 gitsign-credential-cache
+!/cmd/gitsign-credential-cache
```
imjasonh, line 7:
```suggestion
/gitsign-credential-cache
```

This will just exclude the built binary, and not any path matching `*gitsign-credential-cache*`

imjasonh (approved):

.gitignore
wlynch: This will only ignore it if it is in the root directory, which won't exclude binaries built if running `go build .` from the subdirectory itself. 😢

imjasonh: booooooo

cpanato (commented):
lgtm

thanks

Issue #98, opened 2022-07-19
question: GitHub app integration


abhishiekgowtham:
I've started using this recently and need to login to GitHub each time i try to push something to the origin using GitHub's desktop application which gets annoying when dealing with multiple PRs/repos. Is there a way cache or store the login creds for a time period instead?

imjasonh:
There is some work in progress to cache and reuse certificates, here: https://github.com/sigstore/gitsign/tree/main/cmd/gitsign-credential-cache

There are some noteworthy caveats and limitations at this time, but it's actively being worked on by @eddiezane and @wlynch

If you try it out and let us know how it goes that would be helpful feedback!

rawkode:
Would be good to get the credential-cache binary into homebrew and allow it to be enabled with `brew service` rather than using `gitsign-credential-cache &`

imjasonh:
A Chrome extension is an interesting approach. I'd have to think about how it would cache things in a place the `gitsign` (or `cosign`) executables could get to them, but it might work.

Now that I'm thinking of it, what I _really_ want a Chrome extension for is closing the OAuth window after I've finished OAuthing (https://github.com/sigstore/sigstore/pull/484)

abhishiekgowtham:
@imjasonh I'd be happy to give it a try and let you know the feedback. Please let me know what branch/tag version I should be using.

imjasonh:
> @imjasonh I'd be happy to give it a try and let you know the feedback. Please let me know what branch/tag version I should be using.

You can try it out now following instructions at https://github.com/sigstore/gitsign/tree/main/cmd/gitsign-credential-cache, just be aware that it might change or go away in the future, and that there are security implications to caching these credentials.

eddiezane:
I started working on dropping the extra binary and using the OS keystore. If that winds up not working I wanted to bake everything into a single binary that forks the daemon like gopls does. Just need to get a chance to finish it.

PR #121, opened 2022-08-22
Add file based configuration.

### LINKED ISSUE #99: File based configuration
**Description**

It would be great to wire up [viper](https://pkg.go.dev/github.com/spf13/viper) or something similar to have file based configuration for gitsign (i.e. `~/.gitsign`) rather than needing to export a bunch of environment variables.

wlynch:
#### Summary

This adds [git-config](https://git-scm.com/docs/git-config) based configuration for gitsign values.

Caveats:
- GITSIGN_CREDENTIAL_CACHE intentionally not supported since this is
  likely changing soon (see https://github.com/sigstore/gitsign/issues/98)
- We do not use subsections (i.e. we use gitsign.issuer instead of gitsign.oidc.issuer).
  This gives us some wiggle room to add host based config later on
  similar to https://git-scm.com/docs/gitcredentials

Fixes #99

#### Release Note

* Adds [git-config](https://git-scm.com/docs/git-config) based configuration

#### Documentation

Will update once new release is cut. This is a backwards compatible change.

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -1,0 +1,113 @@
+// Copyright 2022 The Sigstore Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package config
+
+import (
+	"os"
+
+	"github.com/go-git/go-git/v5"
+	"github.com/go-git/go-git/v5/config"
+	format "github.com/go-git/go-git/v5/plumbing/format/config"
+)
+
+// Config represents configuration options for gitsign.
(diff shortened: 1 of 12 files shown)


cpanato (approved):
cool

lgtm

imjasonh (approved):

internal/config/config.go
```diff
@@ -0,0 +1,113 @@
...
+			out.Rekor = o.Value
+		case "clientID":
+			out.ClientID = o.Value
+		case "redirectURL":
+			out.RedirectURL = o.Value
+		case "issuer":
+			out.Issuer = o.Value
+		case "logPath":
+			out.LogPath = o.Value
+		}
```
eddiezane, line 103: I know git eats output but Is there a way you can warn about an unknown (or mistyped) value in the default case?

eddiezane (approved):
This looks great! One foot gun preventing suggestion.

internal/config/config.go
wlynch: There's a bit of a chicken and egg problem, since we use the config to set up some of the output writers (i.e. GITSIGN_LOG). 

We could refactor just the tty input/output setup to capture this if we *really* wanted, but I'm also not too worried about it - this behavior is something that is present for most git configs (i.e. you can do `git config --local gpg.foo bar` and it won't complain), and it seems to be manageable for most people. We still have `-v` as a fall back to verify parsed values as a sanity check.

PR #122, opened 2022-08-25
Add connectorID option

### LINKED ISSUE #114: Select auth provider with config
Probably dependent on #99, but something that would be nice is if we had a way for users to pre-select the auth provider they want to use for a repo.

Currently we use the default interactive provider, but if we had a per-repo config, then we could have an option to specify the TokenGetter to use https://github.com/sigstore/sigstore/blob/2ec832bbea880d6fb3f7b0b7cdfaffde490c571e/pkg/oauthflow/flow.go#L68-L73

Would still require a browser window to go through the oauth flow, but would save a click if you already knew what you wanted to use.

wlynch:
#### Summary

Allows users to preselect of auth provider to save a click going through
the auth flow.

Fixes #114 

#### Release Note

Auth provider can be automatically selected with the `connectorID` config option. e.g. to always use Google to sign your commits:

```sh
git config gitsign.connectorID https://accounts.google.com
``` 

Note: Valid values are dependent on the OIDC token provider. For valid values for the public Sigstore instance (`oauth2.sigstore.dev`) see README.md.

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,26 +58,28 @@ $ git config --local gitsign.fulcio https://fulcio.example.com
 
 The following config options are supported:
 
-| Option      | Default                          | Description                                                                                   |
-| ----------- | -------------------------------- | --------------------------------------------------------------------------------------------- |
-| fulcio      | https://fulcio.sigstore.dev      | Address of Fulcio server                                                                      |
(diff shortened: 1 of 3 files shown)


imjasonh (approved):
Nice.

Issue #166, opened 2022-10-17
gitsign timestamp cert verification broken


wlynch:
**Description**

My fault for not catching this pre-submission, but looks like timestamp verification doesn't work if the timestamp cert isn't included in the fulcio root pool. We'll need to modify the CertPool loading to include the system root.

**Version**

https://github.com/sigstore/gitsign/commit/175234a4dcbc7518c4188e70e0e170c2d9c185f4

PR #167, opened 2022-10-19
Refactor git commit verification into its own interface.

wlynch:
#### Summary

This pulls out git commit verification logic (e.g. everything but Rekor lookups) into it's own interface + struct so that it can be more easily configured with user settings.

Bumps s/s dependency to pull in https://github.com/sigstore/sigstore/commit/4f1461d458a2a0a761e60312d271b52f25af42aa

The only new change in behavior is the inclusion of the system cert pool in the default cert pool used for validation.

This is setup for #166 to allow for more flexibility in client creation. (https://github.com/sigstore/gitsign/compare/main...wlynch:certpool if you want to see what this will look like)

#### Release Note

BREAKING CHANGE: `git.Verify` now requires a new `git.Verifier` param.

#### Documentation

NONE

### DIFF
--- a/command_verify.go
+++ b/command_verify.go
@@ -45,12 +45,17 @@ func commandVerify(cfg *config.Config) error {
 		return fmt.Errorf("failed to read signature data (detached: %T): %w", detached, err)
 	}
 
+	cv, err := git.NewCertVerifier()
+	if err != nil {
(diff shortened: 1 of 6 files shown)


cpanato (dismissed):
looks cool

cpanato (approved):

PR #170, opened 2022-10-20
Add fulcio root config option.

### LINKED ISSUE #169: gitsign doesn't respect SIGSTORE_ROOT_FILE
**Description**

There’s feature skew between how [sigstore/sigstore](https://github.com/sigstore/sigstore/blob/v1.4.4/pkg/fulcioroots/fulcioroots.go#L69) implements Fulcio cert loading and how [cosign](https://github.com/sigstore/cosign/blob/30bf1c09c6fde849245648ff294725d86094fe28/internal/pkg/cosign/fulcio/fulcioroots/fulcioroots.go#L59) does it. Because of this gitsign doesn’t respect SIGSTORE_ROOT_FILE , it only works with TUF.

We should look to get parity with cosign here.

**Version**

v0.3.2

wlynch:
#### Summary

Allows callers to specify a Fulcio root to use when verifying
signatures. Useful for configuring gitsign to use non-TUF instances of
Sigstore.

Creates new fulcioroot package to allow configuration of what certs get
loaded in.

Dependent on #167
Fixes #169 

#### Release Note

Adds GITSIGN_FULCIO_ROOT config option.

#### Documentation

README updated, will update docs for next release.

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,11 +82,18 @@ The following config options are supported:
 | GITSIGN_OIDC_REDIRECT_URL   | ✅                 |                                  | OIDC Redirect URL                                                                                                                                                                                                                          |
 | GITSIGN_REKOR_URL           | ✅                 | https://rekor.sigstore.dev       | Address of Rekor server                                                                                                                                                                                                                    |
 | GITSIGN_TIMESTAMP_AUTHORITY | ✅                 |                                  | Optional address of timestamping authority. If set, a trusted timestamp will be included in the signature.                                                                                                                                 |
+| GITSIGN_FULCIO_ROOT         | ✅                 |                                  | Path to PEM encoded certificate for Fulcio CA                                                                                                                                                                                              |
 
 For environment variables that support `Sigstore Prefix`, the values may be
 provided with either a `GITSIGN_` or `SIGSTORE_` prefix - e.g.
 `GITSIGN_CONNECTOR_ID` or `SIGSTORE_CONNECTOR_ID`. If both environment variables
 are set, `GITSIGN_` prefix takes priority.
+
+#### Other environment variables
+
+| Environment Variable      | Description                                                                     |
+| ------------------------- | ------------------------------------------------------------------------------- |
+| SIGSTORE_REKOR_PUBLIC_KEY | This specifies an out of band PEM-encoded public key to use for a custom Rekor. |
 
 ## Usage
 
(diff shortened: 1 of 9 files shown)


README.md
```diff
@@ -82,12 +82,19 @@ The following config options are supported:
...
+| GITSIGN_FULCIO_ROOT         | ✅                 |                                  | Path to PEM encoded certificate for Fulcio CA                                                                                                                                                                                              |
 
 For environment variables that support `Sigstore Prefix`, the values may be
 provided with either a `GITSIGN_` or `SIGSTORE_` prefix - e.g.
 `GITSIGN_CONNECTOR_ID` or `SIGSTORE_CONNECTOR_ID`. If both environment variables
 are set, `GITSIGN_` prefix takes priority.
 
+#### Other environment variables
+
+| Environment Variable      | Description                                                                     |
```
znewman01, line 94: Do we want to mention `SIGSTORE_FULCIO_ROOT` here too

README.md
```diff
@@ -141,6 +148,33 @@ Validated Git signature: true
 Validated Rekor entry: true
 ```
 
+### Private Sigstore
+
+[Similar to cosign](https://docs.sigstore.dev/cosign/custom_components/),
+Gitsign can also run against other Sigstore instances besides the default public
+instance.
+
+There are three options to configure Gitsign to verify against custom
```
znewman01, line 157: Is there a reason to duplicate the content from the link rather than just say "see the link for how to do this"?

It feels like an Advanced Use Case (TM) and detracts from the succinctness of the current README

(won't push you super hard on this though)

znewman01 (dismissed):
So sorry you have to make your own fulcioroots.go :'(

Hopefully that will change long-term.

README.md
wlynch: Technically covered above since it's an alias to GITSIGN_FULCIO_ROOT.

SIGSTORE_REKOR_PUBLIC_KEY is special in that it's embedded directly in the cosign library so gitsign doesn't have control over it 😭

README.md
wlynch: Done!

wlynch:
> So sorry you have to make your own fulcioroots.go :'(
>
> Hopefully that will change long-term.

I intentionally wrote it fairly generic so it should be easy to steal for sigstore-go when we're ready 😎

znewman01 (approved):

PR #191, opened 2022-11-16
Add `gitsign show` subcommand.

### LINKED ISSUE #105: proposal: provenance generation and verifying attestation using cosign with new predicate type
**TL;DR:** _Create a provenance by traversing each commit. Attest into OCI registry. Verify all commits in attestation using cosign._

---

**Abstract**

This issue proposes introducing new `cosign.sigstore.dev/attestation/gitsign/v1` predicate type for using the [in-toto/attestation model](https://github.com/in-toto/attestation). Also proposes a new subcommand in gitsign itself: `attestation generate` to generate an attestation file by traversing _each commits_.

**Motivation**

While working on the Cosign VULN_SPEC proposal a few months ago, we learned a lot from the both sigstore and in-toto community members and valuable comments. And no doubt this proposal will be our next opportunity for new learnings for sure. Gitsign is a brand-new tool and very promising. So with @developer-guy, we thought that this proposal would be worth creating and discussing further here. So created this one, as far as we can do best.

**Proposal**

The main idea is to generate a provenance for the entire commit history for the repository in the container image build-time (to prevent a slight time window in which it could have tampered). And storing as an attestation format in the OCI registry. As a consumer of the container image, verifying _gitsign attestation_ means that I verify the _attestation itself_ and each commit that has been made by committers using their public key that is either fetched from _Rekor_ (PKCS7 cert) or GPG verify (PGP cert).

By doing so; rather than saying _"I trust this repository and binaries thus all artifacts are signed"_, the statement becomes _"all commits that have made by commiters are signed & verified by either using gitsign and GPG, so I trust each commit. and also container image, binaries and artifacts already signed."_.

And example use-case for verifying is to use cosign to verify attestation and also verify all the commits by using gitsign under the hood.

Whis proposal requires collaborate with gitsign and cosign together. Implement all of this, we would do add some commands:

For gitsign: introduce new `attestation generate` command:

```
$ gitsign attestation generate
```

For cosign: introduce new `--gitsign true|false` flag:

```
$ COSIGN_EXPERIMENTAL=1 cosign verify-attestation \
    --type cosign.sigstore.dev/attestation/gitsign/v1 \
    --gitsign \
    foo/bar:baz
```

New `--gitsign` flag traverses each commits and verifies by running under the hood:
```
$ cosign verify-blob \
  --cert commits[*].rekor.signature.publicKey \
  --signature commits[*].rekor.signature.content \
  commits[*].author.digest.sha1
```

**The 30,000-foot view**

Please note that this spec should be generic and open for extendibility. This is not intended to be specially built for just gitsign. Wnyone who wants to create their own git commit signing tool should be able to consume this. Here is how new `cosign.sigstore.dev/attestation/gitsign/v1` predicate looks like (draft - open for feedback):

```json
{
    "_type": "https://in-toto.io/Statement/v0.1",
    "subject": [],
    "predicateType": "cosign.sigstore.dev/attestation/gitsign/v1",
    "predicate": {
        "invocation": {
            "fulcio_url": "https://fulcio.sigstore.dev",
            "rekor_url": "https://rekor.sigstore.dev",
            "oidc": {
                "client_id": "sigstore",
                "issuer": "https://oauth2.sigstore.dev/auth",
                "redirect_url": ""
            },
            "timestamp": 1627564731
        },
        "commits": [
            {
                "issuer": {
                    "o": "sigstore.dev",
                    "cn": "sigstore-intermediate"
                },
                "signature": {
                    "status": "G",
                    "format": "pkcs7",
                    "publicKey":"base64-encoded-pkcs7-cert"
                },
                "rekor": {
                    "signature": {
                        "algorithm": "ecdsa-with-SHA256",
                        "content": "MEYCIQC1gffaJyVGfmJNoX94n9vOj+1EKEAlolT9UH7Bb2MuwwIhAJ1FFee9bDQdLbjt4yjbYz5Ojd3uITilNU4KLGJqiSQr",
                        "cert": "base64-encoded-cert"
                    },
                    "digest": {
                        "sha256": "d00bf6f1a50835a372b137ea18306d6a1d554500caf821e375817173db760868"
                    }
                },
                "author": {
                    "name": "foo@test.com",                 
                    "digest": {
                        "sha1": "6a2b1f12938059f2fbe68754657fad33a4fca372"
                    }
                }
            },
            {
                "signature": {
                    "status": "G",
                    "format": "pgp",
                    "publicKey": "base64-encoded-pgp-cert"
                },
                "author": {
                    "name": "bar@test.com",
                    "digest": {
                        "sha1": "5ee8eca4f05ef4b79eb6dc21c09198d4d61e3495"
                    }
                }
            },
            {
                "signature": {
                    "status": "N"
                },
                "author": {
                    "name": "baz@test.com",
                    "digest": {
                        "sha1": "adba2b44bd04f8f2c92fbf739af797b8dcd046b5"
                    }
                }
            }
        ]
    }
}
```

* author.name: (required) name of the author
* author.digest.sha1: (required) commit

* signature.status: (required) [`%G?`](https://git-scm.com/docs/git-log#Documentation/git-log.txt-emGem) of a commit
* signature.format: (optional) signed with? (gpg, gitsign, etc.)
* signature.publiKey: (optional) certificate from ($ git cat-file commit HEAD)

* rekor: (optional) (if keyless signed with gitsign)
* rekor.signature.algorithm: (required) signature algorithm
* rekor.signature.content: (required) Rekor.Body.HashedRekordObj.signature.content
* rekor.signature.cert: (required) Rekor.Body.HashedRekordObj.signature.publicKey.content

* rekor.digest.sha256: (required) Rekor.Body.HashedRekordObj.data.hash.value

* issuer: (optional) (if keyless signed)
* issuer.o: (required) organization 
* issuer.cn (required) commonName

**Implementation**

1. Define a new [SLSA provenance predicate type](https://github.com/in-toto/attestation/tree/main/spec) either here in gitsign or in the in-toto repository.

2. We should add a new `attestation generate` command in gitsign:

```
$ gitsign attestation generate -o gitsign.json
```

See [git-log(1)](https://git-scm.com/docs/git-log) page for more details about the [`%G?`](https://git-scm.com/docs/git-log#Documentation/git-log.txt-emGem) signature status.

Pseudocode:
```
loop through each commits
  check %G? for each commit
    if commit no signed: // <sig>N</sig>
      set signature.status = N
    if commit signed with GPG: // <sig>G</sig>
      set signature.status = G
      set signature.format = GPG
      set signature.publicKey = base64-encoded-pgp-cert
    if commit signed with gitsign:
      search on Rekor: rekor-cli search --artifact <COMMIT>
        get certificate: $(rekor-cli get --uuid=$uuid --format=json | jq -r .Body.HashedRekordObj.signature.publicKey.content)
        get signature: $(rekor-cli get --uuid=$uuid --format=json | jq -r .Body.HashedRekordObj.signature.content)
        get issuer from the cert
        get necessary rekor obj values from rekor API
        append issuer and rekor objects
    append author object
```

3. Introduce a new `--gitsign` boolean flag in `cosign verify-attestation`:

```
$ COSIGN_EXPERIMENTAL=1 cosign verify-attestation \
    --type cosign.sigstore.dev/attestation/gitsign/v1 \
    --gitsign \
    foo/bar:baz
```

Cosign should first verify the attestation as is. If verification succeed, it should parse the `.att` to load all `cosign.sigstore.dev/attestation/gitsign/v1` predicate type to struct definitions.

The New `--gitsign` flag traverses each commit and verifies by running under the hood:
```
$ cosign verify-blob \
  --cert commits[*].rekor.signature.publicKey \
  --signature commits[*].rekor.signature.content \
  commits[*].author.digest.sha1
```

Pseudocode:
```
loop through each commits
  verify the attestation
    if verification fails
      exit
    if verification succeed
      parse predicate to structs
        traverse in commits[*]
          if commit signed with gitsign
            run $ cosign verify-blob --cert .rekor.signature.publicKey --signature .rekor.signature.content .author.digest.sha1
          if commit signed with gpg
            run $ git verify-commit .author.digest.sha1 (in cosign)
          if commit unsigned
            exit
          if one of verification fails exit
```

**End User Usage**

1. Generates a provenance:
```
$ gitsign attestation generate -o gitsign.json
```

2. Attests and push to OCI:
```
$ COSIGN_EXPERIMENTAL=1 cosign attest \
    --type cosign.sigstore.dev/attestation/gitsign/v1 \
    --predicate gitsign.json \
    foo/bar:baz
```

3. Verifies the attestation:
```
$ COSIGN_EXPERIMENTAL=1 cosign verify-attestation \
    --type cosign.sigstore.dev/attestation/gitsign/v1 \
    --gitsign \
    foo/bar:baz
```

**Intended Users**
* container image builders
* pipelines
* end users

**User Story**

This container image seems signed by a trusted author but still, I couldn't trust the source code because I don't know who contributed. There is always the possibility that committers could have impersonated someone else by using their name and email. What I want is to verify all the commits that include in the source code that makes up the final container image. So I can be sure that there is no _unsigned_ commit. All commits have a signature and are verified.

**Concerns**

1. We should be careful here since our main goal here is to not just say _"all commits are verified"_ but _"all commits are verified and trusted"_. So we should not forget that.

2. Backward compatibility is another big issue for us. What if a repository started enforcing GPG after the 50th commit and switched to a brand-new gitsign tool after the 200th commit? How the whole concept will work here?

Would pass something `-n | --number <INTEGER>` flag to `gitsign attestation generate` tackle this problem? So we can only get commits after _n_ th.

**Alternative Methods**

1. Instead of generating an attestation and verifying, we can simply enforce _signed commits only_ rule in the Git providers (GitHub, GitLab, etc.). So It would be easier to reject _unsigned_ commits. E.g. server-side precommit hooks. 

2. Similar to 1, instead of trusting the commits, we can trust the commit authors. By making an allowlist for the public keys that committers use, we can create restrictive policies to reject upcoming commits from the anonymous (i.e. open-source) contributors.

**Related Proposals**

1. XREF: https://github.com/sigstore/gitsign/issues/94 by @wlynch

What distinguishes our proposal from the following above is that we do not create attestation on each commit, but instead keep a single one. Also this proposal does NOT store _anything_ in `refs` folder. We also want to verify everything as much as possible before using.

**Open Questions**

1. Does this all make sense overall?
2. Are all fields in `rekor` object really necessary? How would it be if we just pass `UUID` instead?
3. Should we verify each commit in history that made up the container image? (before running the container image)
4. Instead of generating the attestation that only has one signature, shouldn't each commiter sign the same attestation with their PK or keyless somehow? 
5. Should we sign all commits individually?
6. Should we store the final attestation in a version controlled in the repository itself? (i.e. in ref/, so anyone can verify all commits using gitsign itself either)
7. Does this proposal uses Zero Trust ("trust nothing, verify everything") principles?

Waiting for your feedback!

cc @dlorenc @lukehinds @TomHennen @adityasaky @SantiagoTorres

wlynch:
#### Summary

Adds a new subcommand that outputs an in-toto Statement containing the source provenance of the given commit. This works for both Gitsign + GPG signed git commits - it extracts the cert information if present.

This can be coupled with other tools like cosign to produce attestations, i.e.:

```sh
$ cosign attest --predicate <(gitsign show) <image>
```

Also moves io Stream creation into the root RunE func so that streams aren't created for subcommands that don't need them.

Possibly fixes #105 (@Dentrax @developer-guy let me know what you think!)

cc @adityasaky (you may be interested in this)

#### Release Note

* 🎁 Adds new `gitsign show` subcommand to print out an in-toto Statement for a given Git commit.

#### Documentation

I plan on generating cobra docs in another PR!


nsmith5 (commented):
Might have just missed this on my reading, but can we add the same verification logic from `gitsign --verify` as well when we parse the signature to check its authentic?

wlynch:
> Might have just missed this on my reading, but can we add the same verification logic from `gitsign --verify` as well when we parse the signature to check its authentic?

I'm not against it, but the nice thing about just formatting without verification is that this works for any signature type - GPG, SSH, x509, Gitsign, etc.

If we add verification then we'd probably need to add support for the other key types to stay consistent.

nsmith5:
Ah I see, is there a generic `git` command someone could run first to do verification prior to this that works for any signature type?

wlynch:
> Ah I see, is there a generic `git` command someone could run first to do verification prior to this that works for any signature type?

Not really 😬 There's `git verify-commit` but that just shells out to whatever signing tool is available (so if you have gitsign configured it can only verify gitsign). Git itself assumes that everything in the repo uses the same signing scheme (which is unfortunate since that's not what users do in practice in a shared repo).

What we can do as a happy medium is add a `--verify` flag, though I'm not sure if we should make this default to true or false. 🤔

nsmith5:
I'd prefer `--verify true` as the default. I feel like `--verify false` default makes it easy to add an attestation without actually checking its authentic. I'd even be in favour of the flag being someing like `--insecure-skip-verification` to make it clear that you're just passing the signature along without checking. And maybe also state that in the attestation to make it clear that we didn't actually check the signature so like

```diff
{
  "_type": "https://in-toto.io/Statement/v0.1",
  "predicateType": "gitsign.sigstore.dev/predicate/git/v0.1",
  "subject": [
    {
      "name": "git@github.com:wlynch/gitsign.git",
      "digest": {
        "sha1": "10a3086104c5331623be85a5e30d709f457b536b"
      }
    }
  ],
  "predicate": {
    "source": {
      "tree": "194fca354a2439028e347ce5e19e4db45bd708a6",
      "parents": [
        "2eaf8fc6d66505baa90640d018e1131cd8e99334"
      ],
      "author": {
        "name": "Billy Lynch",
        "email": "billy@chainguard.dev",
        "date": "2022-11-14T16:13:19-05:00"
      },
      "committer": {
        "name": "Billy Lynch",
        "email": "billy@chainguard.dev",
        "date": "2022-11-14T16:13:19-05:00"
      },
      "message": "add sample\n"
    },
+  "signature_verified": false,
    "signature": "-----BEGIN SIGNED MESSAGE-----\nMIIEAwYJKoZIhvcNAQcCoIID9DCCA/ACAQExDTALBglghkgBZQMEAgEwCwYJKoZI\nhvcNAQcBoIICpDCCAqAwggImoAMCAQICFFTzLmXKAlKX5xTUaYoUE5giCxZvMAoG\nCCqGSM49BAMDMDcxFTATBgNVBAoTDHNpZ3N0b3JlLmRldjEeMBwGA1UEAxMVc2ln\nc3RvcmUtaW50ZXJtZWRpYXRlMB4XDTIyMTExNDIxMTMyM1oXDTIyMTExNDIxMjMy\nM1owADBZMBMGByqGSM49AgEGCCqGSM49AwEHA0IABH++UCDlF9MaQCSgDKQ0bWhD\neOmTrk1sEHw9Oel1eCyrr3SFhDAghcO3VwO7baYmL16fUwRYwMhj5urowsLVrjKj\nggFFMIIBQTAOBgNVHQ8BAf8EBAMCB4AwEwYDVR0lBAwwCgYIKwYBBQUHAwMwHQYD\nVR0OBBYEFHMPDOs6IDY/iRnVqacIj/yvJbNpMB8GA1UdIwQYMBaAFN/T6c9WJBGW\n+ajY6ShVosYuGGQ/MCIGA1UdEQEB/wQYMBaBFGJpbGx5QGNoYWluZ3VhcmQuZGV2\nMCkGCisGAQQBg78wAQEEG2h0dHBzOi8vYWNjb3VudHMuZ29vZ2xlLmNvbTCBigYK\nKwYBBAHWeQIEAgR8BHoAeAB2AN09MGrGxxEyYxkeHJlnNwKiSl643jyt/4eKcoAv\nKe6OAAABhHf9WZAAAAQDAEcwRQIgV8anMDEjbHI/WvGxpJmm44DgBTYf5bkfBJIP\n6FJtqXYCIQD/noLzthDKgjrXoiep/BqqnygoTRM9HKim+DRMbwHteDAKBggqhkjO\nPQQDAwNoADBlAjEAvHvqOAKT34QQx9PSuOswQfquByALdzA1ES0nx4M5i47kqNeE\nBl612/hYTD1ydpLIAjBTWiHDtdxM9rriTIyGGJubC0+vNcccsURDTJ+A3XnMAER3\nikl/cJ2wG9c8ZN7AUS8xggElMIIBIQIBATBPMDcxFTATBgNVBAoTDHNpZ3N0b3Jl\nLmRldjEeMBwGA1UEAxMVc2lnc3RvcmUtaW50ZXJtZWRpYXRlAhRU8y5lygJSl+cU\n1GmKFBOYIgsWbzALBglghkgBZQMEAgGgaTAYBgkqhkiG9w0BCQMxCwYJKoZIhvcN\nAQcBMBwGCSqGSIb3DQEJBTEPFw0yMjExMTQyMTEzMjNaMC8GCSqGSIb3DQEJBDEi\nBCC9Yk93XCRKy6FPCb8dAqjdWpjb1NIbFtTo9CP6yYOZQjAKBggqhkjOPQQDAgRH\nMEUCIQCq+2Zs0bBcAAciePeeRpzmfVJ2gEu7sGngy+TcYpS0ugIgL9Qix3V8taBV\n+Tb6rMZmt80sfGsYhUqE8KsIF1AEc+8=\n-----END SIGNED MESSAGE-----\n",
    "signer_info": [
      {
        "attributes": "MWkwGAYJKoZIhvcNAQkDMQsGCSqGSIb3DQEHATAcBgkqhkiG9w0BCQUxDxcNMjIxMTE0MjExMzIzWjAvBgkqhkiG9w0BCQQxIgQgvWJPd1wkSsuhTwm/HQKo3VqY29TSGxbU6PQj+smDmUI=",
        "certificate": "-----BEGIN CERTIFICATE-----\nMIICoDCCAiagAwIBAgIUVPMuZcoCUpfnFNRpihQTmCILFm8wCgYIKoZIzj0EAwMw\nNzEVMBMGA1UEChMMc2lnc3RvcmUuZGV2MR4wHAYDVQQDExVzaWdzdG9yZS1pbnRl\ncm1lZGlhdGUwHhcNMjIxMTE0MjExMzIzWhcNMjIxMTE0MjEyMzIzWjAAMFkwEwYH\nKoZIzj0CAQYIKoZIzj0DAQcDQgAEf75QIOUX0xpAJKAMpDRtaEN46ZOuTWwQfD05\n6XV4LKuvdIWEMCCFw7dXA7ttpiYvXp9TBFjAyGPm6ujCwtWuMqOCAUUwggFBMA4G\nA1UdDwEB/wQEAwIHgDATBgNVHSUEDDAKBggrBgEFBQcDAzAdBgNVHQ4EFgQUcw8M\n6zogNj+JGdWppwiP/K8ls2kwHwYDVR0jBBgwFoAU39Ppz1YkEZb5qNjpKFWixi4Y\nZD8wIgYDVR0RAQH/BBgwFoEUYmlsbHlAY2hhaW5ndWFyZC5kZXYwKQYKKwYBBAGD\nvzABAQQbaHR0cHM6Ly9hY2NvdW50cy5nb29nbGUuY29tMIGKBgorBgEEAdZ5AgQC\nBHwEegB4AHYA3T0wasbHETJjGR4cmWc3AqJKXrjePK3/h4pygC8p7o4AAAGEd/1Z\nkAAABAMARzBFAiBXxqcwMSNscj9a8bGkmabjgOAFNh/luR8Ekg/oUm2pdgIhAP+e\ngvO2EMqCOteiJ6n8GqqfKChNEz0cqKb4NExvAe14MAoGCCqGSM49BAMDA2gAMGUC\nMQC8e+o4ApPfhBDH09K46zBB+q4HIAt3MDURLSfHgzmLjuSo14QGXrXb+FhMPXJ2\nksgCMFNaIcO13Ez2uuJMjIYYm5sLT681xxyxRENMn4DdecwARHeKSX9wnbAb1zxk\n3sBRLw==\n-----END CERTIFICATE-----\n"
      }
    ]
  }
}
```

My thinking here is that folks need to make policy decisions based on this attestations and might not have access to go pull the source and verify themselves at that point (that is why we're using this attestation in the first place) so we want some clear signals that it was verified by the entity that signed this attestation.

nsmith5:
If we just fail out on SSH signatures and x509 is ok for now? I feel like if you're reaching for this tool its a corner case because you probably

- Have no sig
- Got signed by github GPG key
- Have a gitsign signature

So we'll be able to handle the common situations

internal/commands/show/show.go
```diff
@@ -0,0 +1,217 @@
...
+		}
+		predicate.SignerInfo = sigs
+	}
+
+	// Try and resolve the remote name to use as the subject name.
+	// If the repo does not have a remote configured then this will be left
+	// blank.
+	if remote == "" {
+		remote = "origin"
+	}
```
Dentrax, line 127: I think cobra already sets a default value at `StringVarP` func.

internal/commands/show/show.go
```diff
@@ -0,0 +1,217 @@
...
+				Email: commit.Committer.Email,
+				Date:  commit.Committer.When,
+			},
+			Message: commit.Message,
+		},
+		Signature: commit.PGPSignature,
+	}
+
+	// If we can a PEM encoded signature, try and extract certificate details.
+	pem, _ := pem.Decode([]byte(commit.PGPSignature))
```
Dentrax, line 113: Maybe better not to ignore `err` checking and `pem.Bytes > 0` for more strict error checking?

wlynch:
> If we just fail out on SSH signatures and x509 is ok for now? I feel like if you're reaching for this tool its a corner case because you probably
> 
> * Have no sig
> * Got signed by github GPG key
> * Have a gitsign signature
> 
> So we'll be able to handle the common situations

We wouldn't be able to verify anything that isn't Gitsign, including GPG / GitHub web-flow signed commits which is why I'm worried. 😬

Unsigned commits is also an interesting case - my gut feeling is that I think it's okay to produce provenance documents that don't have a signature. It's up to the attestor to determine if that's something they are willing to sign off on.

internal/commands/show/show.go
wlynch: Actually not an error! It's the left over bytes found during PEM decoding. https://pkg.go.dev/encoding/pem#Decode

internal/commands/show/show.go
wlynch: Removed.

internal/commands/show/show.go
```diff
@@ -0,0 +1,214 @@
...
+const (
+	predicateType = "gitsign.sigstore.dev/predicate/git/v0.1"
+)
+
+type options struct {
+	FlagRemote string
+}
+
+func (o *options) AddFlags(cmd *cobra.Command) {
+	cmd.Flags().StringVarP(&o.FlagRemote, "remote", "r", "origin", "make a signature")
```
eddiezane, line 45:
```suggestion
	cmd.Flags().StringVarP(&o.FlagRemote, "remote", "r", "origin", "specify the git remote")
```

internal/commands/show/show.go
```diff
@@ -0,0 +1,214 @@
...
+
+	// Try and resolve the remote name to use as the subject name.
+	// If the repo does not have a remote configured then this will be left
+	// blank.
+	resolvedRemote, err := repo.Remote(remote)
+	if err != nil && !errors.Is(err, git.ErrRemoteNotFound) {
+		return nil, err
+	}
+	remoteName := ""
+	if resolvedRemote != nil && len(resolvedRemote.Config().URLs) > 0 {
```
eddiezane, line 130: Potential panic here?

internal/commands/show/show.go
```diff
@@ -0,0 +1,214 @@
...
+	cmd := &cobra.Command{
+		Use:   "show [revision]",
+		Short: "Show source predicate information",
+		Long: `Show source predicate information
+
+Prints an in-toto style predicate for the specified revision.
+If no revision is specified, HEAD is used.
+
+This command is experimental, and its CLI surface may change.`,
+		Args: cobra.NoArgs,
```
eddiezane, line 206: This will break if you pass a revision

internal/commands/show/show.go
```diff
@@ -0,0 +1,214 @@
...
+	})
+	if err != nil {
+		return err
+	}
+	revision := "HEAD"
+	if len(args) > 0 {
+		revision = args[0]
+	}
+
+	out, err := statement(repo, o.FlagRemote, revision)
```
eddiezane, line 60: I am confused if `-r` is for a revision or a remote

eddiezane (commented):
Nice job! I have a few things.

internal/commands/show/show.go
eddiezane: Derp nvm.

internal/commands/show/show.go
wlynch: Yeah this was accidentally left in place when I was trying to debug the tty weirdness. Removed.

eddiezane (approved):
/lgtm

PR #192, opened 2022-11-17
Generate CLI docs.

wlynch:
#### Summary

Generate CLI docs for all cobra commands.

#### Release Note

NONE

#### Documentation

📖 📖 📖


imjasonh:
From experience, you may want a CI check that fails if these docs haven't been updated.

`ko` has this: https://github.com/ko-build/ko/blob/main/hack/presubmit.sh#L29-L32

main.go
```diff
@@ -35,8 +33,6 @@ func main() {
 	}
 
 	rootCmd := root.New(cfg)
-	rootCmd.AddCommand(version.New(cfg))
```
eddiezane, line 38: Any reason for this other than to keep main clean?

eddiezane (approved):
+1 to presubmit

main.go
wlynch: Yeah - the doc generator needs the subcommands need to be added to pick them up. This moves the AddCommand's to the rootCmd creation to guarantee both main's have the same command tree.

eddiezane (approved):

PR #196, opened 2022-11-28
Fix Timestamp Authority verification

### LINKED ISSUE #22: Add RFC3161 timestamps to signatures?
**Description**

[We should be able to add RFC3161 signed timestamps into signatures pretty easily](https://pkg.go.dev/github.com/github/smimesign/ietf-cms#SignedData.AddTimestamps).

Rekor does provide a timestamping authority today, but it's currently going through some changes that we would need to track, and potentially we might need to use another TSA altogether - https://github.com/sigstore/rekor/issues/812

cc @nsmith5 @haydentherapper @dlorenc Any thoughts here? Is this something worth adding or is it not worth it?

wlynch:
Part of #166 (e2e test to be re-enabled in another PR - this needs to land first)
Fixes #22

#### Summary

This fixes signing commits with a timestamp authority. To do this we forked the ietf-cms library from smimesign in order to do the following:

- Fixed timestamp before/after checks when the code signing cert issue time matches the signed timestamp. This check should be inclusive. Keyless signing is particularly susceptible to this since we issue certs on the fly.
- Allowed for configurable cert pools for TSA and commit verification. Previously smimesign assumed these would be one in the same, based on the system pool. This allows for different verification options to be configued in order to prevent letting the TSA certs verify the commit signature cert.
- Adds GITSIGN_TIMESTAMP_CERT option to allow loading in of TSA certs.
- Renames GITSIGN_TIMESTAMP_AUTHORITY to GITSIGN_TIMESTAMP_URL.

We're choosing to fork here since upstream hasn't been responsive to PRs.

#### Release Note

- 🐛 Fixes timestamp authority validation.
- 🎁 Adds GITSIGN_TIMESTAMP_CERT config option.
- 🚨 BREAKING CHANGE: Renames GITSIGN_TIMESTAMP_AUTHORITY to GITSIGN_TIMESTAMP_URL

#### Documentation

✅

### DIFF
--- a/.github/workflows/verify.yml
+++ b/.github/workflows/verify.yml
@@ -22,7 +22,11 @@ jobs:
       - name: Check license headers
         run: |
           set -e
-          addlicense -l apache -c 'The Sigstore Authors' -v -ignore *.yml -ignore *.yaml *
+          addlicense -l apache -c 'The Sigstore Authors' -v \
+            -ignore "*.yml" \
+            -ignore "*.yaml" \
+            -ignore "internal/fork/**" \
+            *
           git diff --exit-code
 
   golangci:
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -34,3 +34,5 @@ issues:
 run:
   issues-exit-code: 1
   timeout: 10m
+  skip-dirs:
+    - internal/fork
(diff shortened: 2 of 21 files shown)


priyawadhwa (approved):

priyawadhwa (approved):

haydentherapper:
@wlynch FYI, we're working on a verification library for 3161 timestamps here - https://github.com/sigstore/timestamp-authority/pull/153

PR #211, opened 2023-01-05
Add interactive flow to credential cache.

### LINKED ISSUE #190: Support users who SSH into a remote server
SSHing into a remote machine to develop is a fairly common workflow.

```
ssh me.corp.biz
git init # and setup gitsign
echo foo > README
git commit -am "just fooing"
```

Unfortunately the UX in this case is somewhat suboptimal, though it has improved greatly. Instead of popping up a browser window like it would if you were `git commit`ing locally, this prints a URL and asks you to open in a browser you have access to, whereupon you OAuth, and copy a code back to the waiting gitsign prompt to continue.

This isn't bad (and again, it's much improved!), but it means quite a few more manual steps for each commit: copy/paste URL, the usual OAuth, copy/paste auth code.

To make this easier, it could be cool to provide a local agent that connects to the remote agent over SSH, and listens for browser-open requests, then opens a (local) browser window, gets the token, then sends it back over that same connection to the remote gitsign instance, where it's exchanged for a cert. Or, maybe, the local agent trades the token for the cert and passes only the cert, I don't know.

In any case, I think most of this code already exists in some form already, between gitsign and the cred cache, it's just about packaging it differently, and configuring the handoffs between the local agent and the remote gitsign, without sacrificing security.

wlynch:
#### Summary

This PR adds the ability to let the credential cache to go through the interactive flow if a credential is not found.

This is mostly refactoring to share the interactive flow between the cache and gitsign directly and prevent import cycles.

This allows for the cache socket to be passed through an SSH session to allow for remote sessions to use the cache to fetch credentials with the local client interactive flow (i.e. you can use your local browser to authenticate and cache credentials for a remote session).

Testing around this is weak at the moment - I want to look into setting up scaffolding to allow easier integration tests against test instances, but this will happen in a different PR.

Fixes #190 

#### Release Note

Credential cache to go through the interactive flow if a credential is not found. When paired with SSH socket forwarding, this can be used to use your local browser to authenticate and cache credentials for a remote session.

#### Documentation

✅


wlynch:
cc @kaniini

kaniini:
Total gamechanger for using gitsign on remote machines.  Thanks so much.

internal/fulcio/identity.go
```diff
@@ -40,8 +40,16 @@ import (
 	"golang.org/x/oauth2"
 )
 
+// PrivateKey defines the [crypto.PrivateKey] interface. This should be true for all PrivateKeys.
+type PrivateKey interface {
```
eddiezane, line 44: Have you considered calling this `KeyPair` instead?

internal/io/streams.go
```diff
@@ -69,7 +70,7 @@ func (s *Streams) Wrap(fn func() error) error {
 	// Log any panics to ttyout, since otherwise they will be lost to os.Stderr.
 	defer func() {
 		if r := recover(); r != nil {
-			fmt.Fprintln(s.TTYOut, r)
+			fmt.Fprintln(s.TTYOut, r, string(debug.Stack()))
```
eddiezane, line 73: Intended to leave this in?

eddiezane (approved):
Looks good!

internal/io/streams.go
wlynch: yes. the recover alone isn't super useful.

PR #220, opened 2023-01-13
Add offline verification

### LINKED ISSUE #219: Offline Verification
**Description**

It would be good to modify gitsign to be able to support offline verification.

1. Remove need for clients to reach out to Rekor if they have the full inclusion proof.
2. Get gitsign off of [non-GA Rekor APIs](https://docs.google.com/document/d/1j_pYTKWmRbsif8eHzRwTg3_7IVlMRwTOzDc3DG4GOT0/edit)

Plan is to include the Rekor log entry fields as ASN.1 unauthenticated attributes in the PKCS7 object. This will also allow us to store the log index / uuid in the commit itself.

wlynch:
#### Summary

This adds the ability for gitsign commits to include rekor log entry details to enable offline verification of commits.

- Adds new unauthenticated attributes corresponding to Rekor log entry values.
- Adds support for offline commit verification along side existing SHA based online verification.
- Adds config option for users to select offline or online storage options. Defaults to existing online behavior to allow users to opt-in, this will be changed to offline after a release.
- Adds e2e test for offline verification.

Fixes #219 

#### Release Note

Adds new `rekorMode` config option to allow configuring signature format to allow for offline verification of commits. This is opt-in for this release. This will become the default in the future.

#### Documentation

✅


wlynch:
Marked as draft for now since we still need to define the Rekor OIDs in the Rekor repo itself, but wanted to throw this up for feedback.

cc @znewman01 @haydentherapper @asraa who might be interested.

docs/verification.md
```diff
@@ -0,0 +1,174 @@
...
+xQhKcJMh2qOgC0xVSdMGPTAfBgNVHSMEGDAWgBRYwB5fkUWlZql6zJChkyLQKsXF
++jAiBgNVHREBAf8EGDAWgRRiaWxseUBjaGFpbmd1YXJkLmRldjAsBgorBgEEAYO/
+MAEBBB5odHRwczovL2dpdGh1Yi5jb20vbG9naW4vb2F1dGgwCgYIKoZIzj0EAwMD
+aAAwZQIxAK++9b/nBfUV4geFSADOgR4+unsh0+SkiublOJ8BihZsTNOUr3fwfW6x
+nPkp+Sy1EwIwdOulWop3oJV/Qo7fau0mlsy0MCm3lBgyxo2lpAaI4gFRxGE2GhpV
+7N+kCoS1A24S
+-----END CERTIFICATE-----
+```
+
+Notice that **the Rekor entry uses the same cert that was used to generate the
```
znewman01, line 168: nit: maybe `s/uses/includes/`?

docs/verification.md
```diff
@@ -0,0 +1,174 @@
...
+sha256(der(sort(system time | commit data | content type)))
+```
+
+The resulting Rekor log entry fields and inclusion proof will be stored in the
+PKCS7 object as unauthenticated (i.e. not included in the cryptographic
+signature) attributes.
+
+```
+unauth_attr:
+  object: undefined (1.3.6.1.4.1.57264.3.8)
```
znewman01, line 20: Nit: maybe worth inlining a table describing these extensions? 

(It's a bit of a lie, but you could also annotate this data too.)

internal/commands/root/sign.go
```diff
@@ -78,12 +78,16 @@ func commandSign(o *options, s *gsio.Streams, args ...string) error {
...
 		Detached:           o.FlagDetachedSignature,
 		TimestampAuthority: o.Config.TimestampURL,
 		Armor:              o.FlagArmor,
 		IncludeCerts:       o.FlagIncludeCerts,
-	})
+	}
+	if o.Config.RekorMode == "offline" {
+		opts.RekorAddr = o.Config.Rekor
+	}
+	sig, cert, tlog, err := git.Sign(ctx, rekor, userIdent, dataBuf.Bytes(), opts)
```
znewman01, line 90: Nit: `tlog` is kinda confusing as a name, maybe `tlogEntry` (same elsewhere too)

znewman01: I know that inscrutability is Idiomatic Go (TM) so you can fight me on this

internal/config/config.go
```diff
@@ -39,6 +39,12 @@ type Config struct {
 
 	// Address of Rekor server
 	Rekor string
+	// Rekor storage mode to operate in. One of [online, offline] (default: online)
+	// online - Commit SHAs are stored in Rekor, requiring online verification for all commit objects.
+	// offline - Hashed commit content is stored in Rekor, with Rekor attributes
+	// necessary for offline verification being stored in the commit itself.
+	// Note: online verification will be deprecated in favor of offline in the future.
+	RekorMode string
```
znewman01, line 47: Nit: I'd usually make this an enum

internal/signature/sign.go
```diff
@@ -60,61 +68,70 @@ type Identity interface {
 
 // Sign signs a given payload for the given identity.
 // The resulting signature and cert used is returned.
-func Sign(ident Identity, body []byte, opts SignOptions) ([]byte, *x509.Certificate, error) {
+func Sign(ctx context.Context, ident Identity, body []byte, opts SignOptions) ([]byte, *x509.Certificate, *models.LogEntryAnon, error) {
```
znewman01, line 71: Idle thought: maybe worth making a struct to wrap the sig+cert+log entry?

pkg/rekor/rekor.go
```diff
@@ -42,9 +45,10 @@ import (
 	"github.com/sigstore/sigstore/pkg/cryptoutils"
 )
 
-// Verifier represents a mechanism to get and verify Rekor entries for the given Git commit.
+// Verifier represents a mechanism to get and verify Rekor entries for the given Git data.
 type Verifier interface {
 	Verify(ctx context.Context, commitSHA string, cert *x509.Certificate) (*models.LogEntryAnon, error)
+	VerifySignature(ctx context.Context, sig []byte) (*models.LogEntryAnon, error)
```
znewman01, line 51: Little more docs to distinguish these might be nice.

znewman01 (commented):
Sorry for the delay here!

LGTM (pending the Rekor OIDs upstream). If everyone wrote docs like this code review would be a lot less onerous, thanks. Some suggestions but take or leave 'em

docs/verification.md
wlynch: I don't want to keep this in sync with Rekor. It's only one value now, so hopefully it's less of a burden.

docs/verification.md
wlynch: Done.

internal/config/config.go
wlynch: I'm not against this, but it feels like it would be a good standalone refactor.

internal/signature/sign.go
wlynch: Done!

.github/workflows/e2e.yaml
```diff
@@ -84,6 +84,28 @@ jobs:
...
+            --certificate-github-workflow-repository=${{ github.repository }} \
+            --certificate-github-workflow-sha=${{ github.sha }} \
+            --certificate-oidc-issuer="https://token.actions.githubusercontent.com" \
+            --certificate-identity="https://github.com/${{ github.workflow_ref }}"
+
+          # Extra debug info
+          git cat-file commit HEAD | sed -n '/BEGIN/, /END/p' | sed 's/^ //g' | sed 's/gpgsig //g' | sed 's/SIGNED MESSAGE/PKCS7/g' | openssl pkcs7 -print -print_certs -text
+      - name: Test Sign and Verify commit - offline verification
+        env:
+          GITSIGN_REKOR_MODE: "offline"
```
znewman01, line 98: Hm, is there any way to double check that validation actually happens offline? If you typo'd `GITSIGN_REKOR_MODE` this test wouldn't check the right thing.

You could turn off network access for the `gitsign verify` commands
https://github.com/orgs/community/discussions/24975

docs/verification.md
```diff
@@ -0,0 +1,201 @@
...
+        02be - 48 46 6e 34 47 4e 63 32-65 38 65 42 78   HFn4GNc2e8eBx
+        02cb - 48 6f 4b 41 6c 56 6f 77-44 77 4a 51 72   HoKAlVowDwJQr
+        02d8 - 34 32 53 50 56 37 64 2f-6e 72 73 47 34   42SPV7d/nrsG4
+        02e5 - 77 49 67 4c 49 73 36 77-2b 59 75 39 42   wIgLIs6w+Yu9B
+        02f2 - 2f 35 2b 73 6b 6e 72 51-65 36 58 33 72   /5+sknrQe6X3r
+        02ff - 68 6e 6b 41 65 6a 6d 76-55 6d 4d 5a 5a   hnkAejmvUmMZZ
+        030c - 69 51 75 4d 53 49 59 3d-0a               iQuMSIY=.
+```
+
+These OIDs are defined by [Rekor](https://github.com/sigstore/rekor) and are
```
znewman01, line 86: nit: just one OID

docs/verification.md
```diff
@@ -0,0 +1,201 @@
...
+
+## Offline Verification
+
+In offline Rekor storage mode Gitsign will store a HashedRekord in Rekor
+corresponding to the commit content.
+
+Unfortunately this is a bit complex to query manually. Roughly this is:
+
+```
+sha256(der(sort(system time | commit data | content type)))
```
znewman01, line 11: "commit data" is what, the output of `git cat-file commit <SHA>`?

What's content type? Any why is the time in there if there's already a commit time?

What's the "sort" for?

More generally, what's the motivation for the design here (esp because you could just use the commit SHA)? Does this format come from Git?

internal/commands/root/sign.go
```diff
@@ -88,18 +88,21 @@ func commandSign(o *options, s *gsio.Streams, args ...string) error {
 		opts.UserName = o.Config.CommitterName
 		opts.UserEmail = o.Config.CommitterEmail
 	}
-	sig, cert, tlog, err := git.Sign(ctx, rekor, userIdent, dataBuf.Bytes(), opts)
+	if o.Config.RekorMode == "offline" {
+		opts.RekorAddr = o.Config.Rekor
```
znewman01, line 92: Hmm, I find this a little confusing. If we're "offline", we provide the "RekorAddr"? That's the opposite of what I'd expect.

IIUC this basically does the _same_ signing process, it just additionally stuffs offline verification materials into the git repo.

internal/signature/sign.go
```diff
@@ -67,12 +75,21 @@ type Identity interface {
...
+}
+
 // Sign signs a given payload for the given identity.
 // The resulting signature and cert used is returned.
-func Sign(ident Identity, body []byte, opts SignOptions) ([]byte, *x509.Certificate, error) {
+func Sign(ctx context.Context, ident Identity, body []byte, opts SignOptions) (*SignResponse, error) {
 	cert, err := ident.Certificate()
 	if err != nil {
-		return nil, nil, fmt.Errorf("failed to get identity certificate: %w", err)
+		return nil, fmt.Errorf("failed to get idenity certificate: %w", err)
```
znewman01, line 92: nit: typo "idenity"

internal/signature/sign.go
```diff
@@ -209,3 +242,39 @@ func matchSAN(cert *x509.Certificate, name, email string) bool {
 
 	return false
 }
+
+func attachRekorLogEntry(ctx context.Context, sd *cms.SignedData, cert *x509.Certificate, addr string) (*models.LogEntryAnon, error) {
```
znewman01, line 246: Nit: maybe `createAndAttachRekorLogEntry`?

Attach makes me thing it already exists

internal/signature/sign.go
```diff
@@ -209,3 +242,39 @@ func matchSAN(cert *x509.Certificate, name, email string) bool {
...
+	rekor, err := rekor.New(addr)
```
znewman01, line 247: Would it be easier to pass in the Rekor client directly? I always prefer to dependency-inject clients.

internal/git/git.go
```diff
@@ -18,45 +18,50 @@ package git
...
 	"github.com/sigstore/gitsign/internal/fulcio"
 	"github.com/sigstore/gitsign/internal/signature"
 	"github.com/sigstore/gitsign/pkg/git"
 	"github.com/sigstore/gitsign/pkg/rekor"
-	"github.com/sigstore/rekor/pkg/generated/models"
 )
 
-func Sign(ctx context.Context, rekor rekor.Writer, ident *fulcio.Identity, data []byte, opts signature.SignOptions) ([]byte, *x509.Certificate, *models.LogEntryAnon, error) {
-	sig, cert, err := signature.Sign(ident, data, opts)
+func Sign(ctx context.Context, rekor rekor.Writer, ident *fulcio.Identity, data []byte, opts signature.SignOptions) (*signature.SignResponse, error) {
```
znewman01, line 29: Would it be better to have a separate `SignOffline`? The fact that most of this method is behind an early return that we expect to be hit most of the time is a little bit off-putting.

internal/git/git.go
```diff
@@ -18,45 +18,50 @@ package git
...
+	resp, err := signature.Sign(ctx, ident, data, opts)
 	if err != nil {
-		return nil, nil, nil, fmt.Errorf("failed to sign message: %w", err)
+		return nil, fmt.Errorf("failed to sign message: %w", err)
 	}
 
+	// We're using offline verification style signing - nothing more to do.
```
znewman01, line 35: Note to self (maybe worth inlining in the comment): nothing more to do *because `signature.Sign` already did it*

docs/verification.md
```diff
@@ -0,0 +1,201 @@
...
+        030c - 69 51 75 4d 53 49 59 3d-0a               iQuMSIY=.
+```
+
+These OIDs are defined by [Rekor](https://github.com/sigstore/rekor) and are
+used during verification to reconstruct the Rekor log entry and verify the
+commit signature.
+
+## Online Verification
+
+In online Rekor storage mode Gitsign will store the Git commit SHA in rekor
```
znewman01, line 92: nit: "rekor" -> "Rekor"?

internal/signature/sign.go
znewman01: You already have a rekor.Client in git.Sign.

internal/commands/root/sign.go
znewman01: Oh, I see.  If we pass `opts.RekorAddr`, then `signature.Sign` will do the new Rekor log stuff. Otherwise, `git.Sign` will use the `rekor` we pass in.

That feels...a little fraught. I'd rather make the signing mode more explicit in the arguments, or even have separate `git.SignOffline` and `git.SignOnline`.

internal/signature/sign.go
```diff
@@ -67,12 +75,21 @@ type Identity interface {
...
+// SignResponse is the response from Sign containing the signature and other related metadata.
+type SignResponse struct {
+	Signature []byte
+	Cert      *x509.Certificate
+	// LogEntry is the Rekor tlog entry from the signing operation.
+	// This is only populated if offline signing mode was used.
+	LogEntry *models.LogEntryAnon
+}
+
 // Sign signs a given payload for the given identity.
```
znewman01, line 87: It also uploads to Rekor if `opts.RekorAddr` is provided!

internal/git/git.go
znewman01: Maybe factoring out all the Rekor stuff into a helper function would be better.

docs/verification.md
znewman01, line 92: I had to read this a few times to really grok the difference. IIUC:

1. "online" (old way): *after* creating the Git signature, do the Rekor upload (of the commit SHA). Verifiers use 
2. "offline" (new way): *while* creating the Git signature, do the Rekor upload (of the new sha256 hashed thingy) and stick the LogEntryAnon in the PKCS#7 signature

I wonder if it would be clearer to structure both sections as:

- what gets uploaded to Rekor
- how we sign
- how we verify

znewman01 (approved):
I wrote a bunch of comments but you should interpret that as rambling to myself in order to figure out what's going on. If you find any suggestions helpful, feel free to address them but my approval is not contingent on any of them.

internal/rekor/oid/pbcompat.go
```diff
@@ -0,0 +1,105 @@
...
+	"encoding/hex"
+	"fmt"
+
+	"github.com/go-openapi/swag"
+	v1 "github.com/sigstore/protobuf-specs/gen/pb-go/common/v1"
+	rekorpb "github.com/sigstore/protobuf-specs/gen/pb-go/rekor/v1"
+	"github.com/sigstore/rekor/pkg/generated/models"
+)
+
+// This file contains helper functions from going to/from Rekor types <-> protobuf-specs.
```
haydentherapper, line 27: nice work on this!

internal/git/git.go
wlynch: Done.

internal/git/git.go
wlynch: Shouldn't be a problem with the refactor.

internal/signature/sign.go
wlynch: Done.

docs/verification.md
wlynch: Done.

priyawadhwa (approved):

PR #223, opened 2023-01-17
Add support for systemd socket activation

### LINKED ISSUE #222: Systemd support for gitsign-credential-cache
It would be nice to have systemd user units to run gitsign-credential-cache service. And even better if it would be socket activated

iavael:
#### Summary
In this PR there are systemd user units for gitsign-credential-cache for a better integration on linux systemd with systemd. Also gitsign-credential-cache was modified for socket activation support.

#### Release Note
Added parameter -s to gitsign-credential-cache for socket activation support

#### Documentation
README was adjusted

Closes #222


contrib/systemd/gitsign-credential-cache.socket
```diff
@@ -0,0 +1,9 @@
+[Unit]
+Description=GitSign credential cache socket
+
+[Socket]
+ListenStream=%C/.sigstore/gitsign/cache.sock
+DirectoryMode=0700
+
+[Install]
+WantedBy=default.target
```
cpanato, line 9: nit: new line

iavael: Fixed

cmd/gitsign-credential-cache/main.go
```diff
@@ -32,6 +33,7 @@ import (
 var (
 	// Action flags
 	versionFlag = getopt.BoolLong("version", 'v', "print the version number")
+	systemdFlag = getopt.BoolLong("socket-activation", 's', "use systemd socket activation")
```
wlynch, line 36: I'd lean towards not assigning a short flag for flags that aren't multi-platform.
```suggestion
	systemdFlag = getopt.Bool("systemd-socket-activation", "use systemd socket activation")
```

cmd/gitsign-credential-cache/main.go
```diff
@@ -47,31 +49,53 @@ func main() {
...
-	if err != nil {
-		log.Fatalf("error getting user cache directory: %v", err)
-	}
+	var connChan = make(chan net.Conn)
+	if *systemdFlag {
+		listeners, err := activation.Listeners()
+		if err != nil {
+			log.Fatalf("error getting systemd listeners: %v", err)
+		}
+		for _, l := range listeners {
```
wlynch, line 58: https://pkg.go.dev/github.com/coreos/go-systemd/activation#Listeners says that some of these listeners may return nil - we probably want to filter those out.

wlynch (changes requested):
Thanks for putting this together! 🙏 

Some comments, primarily to be defensive if this flag is invoked on a non-systemd machine.

wlynch:
cc @eddiezane in case you have any other comments

cmd/gitsign-credential-cache/main.go
iavael: This getopt lib accepts either short-only, or short-and-long options. long-only are not supported.

cmd/gitsign-credential-cache/main.go
iavael: Fixed

cmd/gitsign-credential-cache/main.go
```diff
@@ -47,31 +49,53 @@ func main() {
...
-	path := filepath.Join(dir, "cache.sock")
-	if _, err := os.Stat(path); err == nil {
-		os.Remove(path)
-	}
-	fmt.Println(path)
+		path := filepath.Join(dir, "cache.sock")
+		if _, err := os.Stat(path); err == nil {
+			os.Remove(path)
+		}
+		fmt.Println(path)
```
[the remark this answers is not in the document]
iavael: This is kinda pointless: sockets are configured in systemd unit, program gets only file descriptors. And if you want to know what's current socket configuration, then you can just run `systemctl --user status gitsign-credential-cache.socket` and get something like this
```
gitsign-credential-cache.socket - GitSign credential cache socket
     Loaded: loaded (/home/iavael/.config/systemd/user/gitsign-credential-cache.socket; enabled; preset: disabled)
     Active: active (running) since Tue 2023-01-17 17:50:53 +04; 1 day 9h ago
      Until: Tue 2023-01-17 17:50:53 +04; 1 day 9h ago
   Triggers: ● gitsign-credential-cache.service
     Listen: /home/iavael/.cache/sigstore/gitsign/cache.sock (Stream)
     CGroup: /user.slice/user-1000.slice/user@1000.service/app.slice/gitsign-credential-cache.socket

янв 17 17:50:53 workstation systemd[5417]: Listening on gitsign-credential-cache.socket - GitSign credential cache socket.
```
In "Listen" field there will be all listening sockets

But I added printing them just in case.

cmd/gitsign-credential-cache/main.go
```diff
@@ -47,31 +49,53 @@ func main() {
...
+		if err != nil {
+			log.Fatalf("error opening socket: %v", err)
+		}
+		go connToChan(l, connChan)
 	}
 	srv := rpc.NewServer()
 	if err := srv.Register(service.NewService()); err != nil {
 		log.Fatalf("error registering RPC service: %v", err)
 	}
+	for conn := range connChan {
```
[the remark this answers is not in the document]
iavael: Added couple of runtime checks:
* service was started via systemd socket activation
* there are actually some stream listen fd-s passed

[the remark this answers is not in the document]
wlynch: Looks great. Thanks!

cmd/gitsign-credential-cache/main.go
wlynch: ack - I'll replace this with cobra in the near term then 😅 (but in another PR, so no further action needed here).

wlynch (approved):

PR #246, opened 2023-02-21
Add support for checking cert email against user config before signing.

### LINKED ISSUE #104: Verify: require cert to match committer?
This could be a disruptive change to users, so looking for feedback before we commit to this - should we require by default that the cert contained in the git commit match the email of the git committer?

Right now we just check for a valid signature, and users would need to add additional checks to verify the identity matches.

This would be in line with [Sigstore clients should require a provided identity](https://docs.google.com/document/d/1o8_bXIygufgiohJGlmBzqF4_BnXCTfgh4ILgJFJxYRs/edit?resourcekey=0-YEar3v67uoT31kj83dCVvA#) by @haydentherapper and @znewman01, since we don't have an easy way to pass additional params via `git verify-commit` or `git log`. This also means we don't have a good way to specify the OIDC provider or Sigstore instance, so we may need additional out of band config (i.e. envvar / https://github.com/sigstore/gitsign/issues/99) for this to be complete.

https://github.com/sigstore/gitsign/issues/65 may be a hard requirement here, since otherwise we won't be able to generate privacy preserving certs. We could also provide an option to disable this behavior (though prefer not to do this if possible).

Thoughts? cc @imjasonh @nsmith5

wlynch:
#### Summary

This change adds a new config option: gitsign.matchCommitter. This option checks whether the certificate fetched matches the user configured email/name.

For human users, this generally means that the SAN email in the cert matches the `user.email` Git config option.

For non-email based identities (e.g.  machine users), the SAN URI can be specified as the user name (since the URI isn't a valid email).

Gitsign requires at least one condition to match for the check to succeed.

This change does *not* enforce any constraints on verification, since this requires additional checking to know what IdP is considered valid.

Fixes #104 

#### Release Note

* Added `gitsign.matchCommitter` config option. This option checks whether the certificate fetched matches the user configured email/name.

#### Documentation

✅

### DIFF
--- a/docs/committer-verification.md
+++ b/docs/committer-verification.md
@@ -1,0 +1,34 @@
+# Committer Verification
+
+Gitsign can be optionally configured to verify that the committer user identity
+matches the git user configuration (i.e. `user.name` and `user.email`)
+
+To enable committer verification, run `git config gitsign.matchCommitter true`.
+
+Committer verification is done by matching the certificate Subject Alternative
+Name (SAN) against the issued Fulcio certificate in the following order:
+
+1. An `EmailAddresses` cert value matches the committer `user.email`. This
+   should be used for most human committer verification.
+2. A `URI` cert value matches the committer `user.name`. This should be used for
+   most automated user committer verification.
+
+In the event that multiple SAN values are provided, verification will succeed if
+at least one value matches.
+
+## Configuring Automated Users
+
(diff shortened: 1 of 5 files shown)


hectorj2f (approved):

cpanato (approved):
cool!! thanks for that!

PR #247, opened 2023-02-24
Bump sigstore cosign to v2, dep and workflows

k4leung4:
#### Summary

Bump sigstore cosign to v2, dep and workflows


wlynch (approved):
Thanks! 🎉

PR #262, opened 2023-03-22
WIP: gitsign verify

wlynch:


priyawadhwa (approved):

cpanato (approved):

.gitignore
```diff
@@ -2,7 +2,7 @@
 *.swp
 dist/*
 .vscode/*
-gitsign
-gitsign-credential-cache
+/gitsign
```
hectorj2f, line 5: I'm wondering if this is intentional, 🤔 .

wlynch: Yup, it is! Before this would match any path that had gitsign in it, so it was ignoring the `internal/gitsign` folder. This changes it to only ignore files named gitsign in the git root dir - the primary purpose for this was to ignore gitsign binaries built via `go build .`.

PR #285, opened 2023-04-11
Update e2e test to use CDN instead of GCS

haydentherapper:
The TUF repo is now accessible via a CDN instead of GCS.


haydentherapper:
cc @wlynch @bobcallaway

wlynch (approved):

PR #292, opened 2023-04-18
Ensure that io writers are properly closed.

### LINKED ISSUE #291: Error Signing Git Commits with Keyless Github OIDC
**Description**
I am experiencing an issue while using gitsign. I encountered the following error: 
`error: gpg failed to sign the data` and 
`fatal: failed to write commit object`.

To debug the issue, I tried to set the GITSIGN_LOG environment variable to log gitsign's output to a file called gitsign.log in my current working directory. I set the GITSIGN_LOG environment variable correctly and verified that it was set by running echo $GITSIGN_LOG in my terminal.

However, even after running gitsign again, the gitsign.log file was not created and I was unable to view its contents using the cat or ls commands.

I would like to know why the log file is not being created and how to resolve the original issue with gitsign. Any help or guidance would be greatly appreciated.

**Version**
v0.6.0

wlynch:
#### Summary

Because the writers were being defered closed in New, this means that some files could be closed before we had the opportunity to write to them. This moves the close to their own func that can be ran in the correct place.

Partially fixes #291 

#### Release Note

Fixed issue where debug log file was incorrectly closed to early.

### DIFF
--- a/internal/commands/root/root.go
+++ b/internal/commands/root/root.go
@@ -65,6 +65,7 @@ func New(cfg *config.Config) *cobra.Command {
 		DisableAutoGenTag: true,
 		RunE: func(cmd *cobra.Command, args []string) error {
 			s := io.New(o.Config.LogPath)
+			defer s.Close()
 			return s.Wrap(func() error {
 				switch {
 				case o.FlagVersion:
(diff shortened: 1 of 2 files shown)


cpanato (approved):
thank you

PR #383, opened 2023-09-29
upgrade to go1.21

cpanato:
#### Summary
- upgrade to go1.21
- also remove the auto approval and just have the auto-merge


cpanato:
ping @wlynch :)

wlynch:
I think if we bump the setup-go action to 1.21 first, we should be able to get the e2e tests working?

PR #465, opened 2024-03-04
Update to use go1.22 and ci udpates

cpanato:
#### Summary
- update go to 1.22
- update goreleaser flags 
- drop cache this is done in the setup-go action step and always use latest cosign


cpanato:
e2e tests will fail

wlynch (approved):

PR #491, opened 2024-04-10
Remove local-user validation.

wlynch:
#### Summary

This flag isn't required, and we're not really using it today. It's usually used to inform the signing tool what key to use, but for keyless signing we don't need this.

Leaving the flag in because we need it defined for git not to error out. We may start using it as a mechanism to support BYO keys in the future.

#### Release Note

* Passing in `--local-user` no longer required for keyless signing.


wlynch:
cc @adityasaky

cpanato (approved):

PR #571, opened 2024-10-13
Handle GeneralName as SAN

### LINKED ISSUE #565: gitsign verify fails with unhandled critical extension
**Description**

gitsign verify fails with `x509: unhandled critical extension` when the signature is issued using a `username` type IDP.

**Version**

v0.10.2

adityasaky:
#### Summary

Fixes #565

For "username" type IDPs, the identity is stored using the GeneralName type, which Go doesn't support all types of values for. So, verification fails with "unhandled critical extension". I borrowed a snippet of code from cosign that addresses this and added it to gitsign.

Note: I've added it to the fork/ietf-cms package because it re-parses the certificates. I first tried to set it at https://github.com/sigstore/gitsign/blob/main/pkg/git/verifier.go#L110-L115 for all certs with `sd.SetCertificates` but this is thrown away when the certs are re-parsed at https://github.com/sigstore/gitsign/blob/main/internal/fork/ietf-cms/verify.go#L49-L52.

#### Release Note

Fixed unhandled critical extension error for certain identity types

#### Documentation

None

### DIFF
--- a/internal/fork/ietf-cms/verify.go
+++ b/internal/fork/ietf-cms/verify.go
@@ -3,9 +3,11 @@ package cms
 import (
 	"bytes"
 	"crypto/x509"
+	"encoding/asn1"
 	"errors"
 
 	"github.com/github/smimesign/ietf-cms/protocol"
+	"github.com/sigstore/sigstore/pkg/cryptoutils"
 )
 
 // Verify verifies the SingerInfos' signatures. Each signature's associated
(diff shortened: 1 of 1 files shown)


internal/fork/ietf-cms/verify.go
```diff
@@ -120,6 +122,23 @@ func (sd *SignedData) verify(econtent []byte, opts x509.VerifyOptions, tsOpts x5
...
+		// This matches https://github.com/sigstore/cosign/blob/a0752eb40b500316ac417baf4926a2c2d99b39b8/pkg/cosign/verify.go#L236-L248
+		if len(cert.UnhandledCriticalExtensions) > 0 {
+			var unhandledExts []asn1.ObjectIdentifier
+			for _, oid := range cert.UnhandledCriticalExtensions {
+				if !oid.Equal(cryptoutils.SANOID) {
+					unhandledExts = append(unhandledExts, oid)
+				}
+			}
+
+			cert.UnhandledCriticalExtensions = unhandledExts
```
wlynch, line 139: This is filtering out the SANOID, but should we be handling it elsewhere?

adityasaky: We're only filtering it out from the list of unhandled extensions. The value is still matched from cert.Extensions. Not rigorous but `gitsign verify <>` correctly handles when given the right identity and errors out when given the wrong identity where before it was complaining about the unhandled extension. I think there's some mismatch in that a parser somewhere is able to load the value into cert.Extensions but somewhere else (the go std lib?) it's marked as an unhandled ext?

wlynch: https://pkg.go.dev/crypto/x509#Certificate
```
	// Users can access these extensions using Extensions and can remove
	// elements from this slice if they believe that they have been
	// handled.
```

👍 LGTM

wlynch (approved):

PR #584, opened 2024-10-21
Fix tlog entry matching

adityasaky:
#### Summary

When gitsign-credential-cache is used, multiple entries share the same public key / signing cert. The rekor search flow was matching with the payload hash and the public key as parameters, but the results are entries which matches _either condition_, meaning tlog entries that used the same key (from the cache) but for other commits may be used during verification.

#### Release Note

Fixed matching of tlog entries during verification

#### Documentation

NONE

### DIFF
--- a/pkg/rekor/rekor.go
+++ b/pkg/rekor/rekor.go
@@ -25,7 +25,6 @@ import (
 	"encoding/pem"
 	"errors"
 	"fmt"
-	"strings"
 
 	"github.com/go-openapi/runtime"
 	"github.com/go-openapi/strfmt"
@@ -110,12 +109,16 @@ func (c *Client) WriteMessage(ctx context.Context, message, signature []byte, ce
 }
 
 func (c *Client) get(ctx context.Context, data []byte, cert *x509.Certificate) (*models.LogEntryAnon, error) {
+	h := sha256.New()
+	h.Write(data)
+	hash := hex.EncodeToString(h.Sum(nil))
+
 	pem, err := cryptoutils.MarshalCertificateToPEM(cert)
 	if err != nil {
 		return nil, err
 	}
 
-	uuids, err := c.findTLogEntriesByPayloadAndPK(ctx, data, pem)
+	uuids, err := c.findTLogEntriesByPK(ctx, pem)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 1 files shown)


pkg/rekor/rekor.go
```diff
@@ -188,9 +189,10 @@ func (c *Client) Verify(ctx context.Context, commitSHA string, cert *x509.Certif
 	return e, cosign.VerifyTLogEntryOffline(ctx, e, c.publicKeys)
 }
 
-// extractCerts is taken from cosign's cmd/cosign/cli/verify/verify_blob.go.
+// extractCertsForHash is taken from cosign's
+// cmd/cosign/cli/verify/verify_blob.go.
 // TODO: Refactor this into a shared lib.
-func extractCerts(e *models.LogEntryAnon) ([]*x509.Certificate, error) {
+func extractCertsForHash(e *models.LogEntryAnon, hash string) ([]*x509.Certificate, error) {
```
wlynch, line 195:
```suggestion
// This function extracts the data hash and certs from a given LogEntryAnon.
func extractData(e *models.LogEntryAnon) (string, []*x509.Certificate, error) {
```

The original intent for this func was to be a generic "get certs from log entry", and make the policy decisions above where there is more context. We should follow a similar pattern here instead of plumbing through the hash and splitting where the hash vs pk policy checks are being done.

pkg/rekor/rekor.go
```diff
@@ -149,16 +152,14 @@ func (c *Client) get(ctx context.Context, data []byte, cert *x509.Certificate) (
...
+// but only filters by the public key used. Earlier, this filtered on both the
+// payload and the public key, but returned uuids of entries that matched
+// either.
+func (c *Client) findTLogEntriesByPK(ctx context.Context, pubKey []byte) (uuids []string, err error) {
 	params := index.NewSearchIndexParamsWithContext(ctx)
 	params.Query = &models.SearchIndex{}
 
-	h := sha256.New()
-	h.Write(payload)
-	params.Query.Hash = fmt.Sprintf("sha256:%s", strings.ToLower(hex.EncodeToString(h.Sum(nil))))
```
wlynch, line 160: (optional): we may want to consider switching to cosign.FindTLogEntriesByPayload if we think there are going to be many cases of the same Pk being used.

wlynch (commented):
Looks good! One small tweak.

pkg/rekor/rekor.go
adityasaky: done!

pkg/rekor/rekor.go
adityasaky: i can take a look tomorrow, I'm in a bit of a brain freeze atm and I'm not sure what we'd do to counter potential sha-1 collisions here.

wlynch: I don't think we have to worry about that too much. we'll still filter by PK+hash, and collisions would need to be generated by the same key within the inclusion window. It's also questionable if you distrust sha1 altogether if you could even trust the signature content being served to you in a sha1 commit. 😅 

I think the long term answer is move away from the dual signing and only support offline verification, though we'll always need to support verification in some form.

wlynch (approved):

PR #596, opened 2024-11-20
README: add warning about internet access

philips:
#### Summary

gitsign differs from GPG in that it requires an internet access on every commit. This breaks the git model of working offline. So, probably worth calling out.

#### Release Note

NONE

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,19 +32,27 @@ Single Repository:
 
 ```sh
 cd /path/to/my/repository
-git config --local commit.gpgsign true  # Sign all commits
-git config --local tag.gpgsign true  # Sign all tags
 git config --local gpg.x509.program gitsign  # Use gitsign for signing
 git config --local gpg.format x509  # gitsign expects x509 args
+
+# Warning: Setting this will make git commit/tag reliant on internet.
+# Alternatively, don't use these settings and add the -S flag instead.
+git config --local commit.gpgsign true  # Sign all commits
+git config --local tag.gpgsign true  # Sign all tags
+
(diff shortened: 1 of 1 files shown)


philips:
@cpanato signed DCO

README.md
```diff
@@ -32,19 +32,27 @@ Single Repository:
...
 
 ```sh
-git config --global commit.gpgsign true  # Sign all commits
-git config --global tag.gpgsign true  # Sign all tags
 git config --global gpg.x509.program gitsign  # Use gitsign for signing
 git config --global gpg.format x509  # gitsign expects x509 args
+
+
+# Warning: Setting this will make git commit/tag reliant on internet.
+# Alternatively, don't use these settings and add the -S flag instead.
```
adityasaky, line 53: The `-S` flag for commits will still rely on the internet, no? I wonder if this message belongs elsewhere more prominently for gitsign use as a whole?

philips: Yeah, explaining that every signed commit now reaches out to the internet might be worth mentioning somewhere. I had setup this setting and then couldn't figure out why commits were happening so slowly.

adityasaky: perhaps adding some text in the signing commits and signing tags sections makes sense? https://github.com/sigstore/gitsign/blob/main/README.md?plain=1#L115

cpanato (approved):
seems ok to me

thanks

README.md
cpanato: will be good to add some clarification in this section as well

Issue #611, opened 2025-01-03
Add features needed to sign SLSA Source Attestations


TomHennen:
**Description**

The SLSA Source Track defines [summary attestations](https://slsa.dev/spec/draft/source-requirements#summary-attestation) as a way of attesting the high level properties of a source revision.

In addition it wants implementors to be able to create other more detailed [provenance attestations](https://slsa.dev/spec/draft/source-requirements#provenance-attestations).

gitsign seems like the perfect tool to be able to sign these attestations but I ran into a couple of problems when I tried signing a summary attestation.

1. The spec wants the subject to specify the git commit using a digest type of `gitCommit` (a [standard in-toto type](https://github.com/in-toto/attestation/blob/main/spec/v1/digest_set.md#gitcommit-gittree-gitblob-gittag).  However gitsign currently stores the git commit in a `sha256` digest type.
2. The spec wants implementors to specify the branches as _annotations_ on the subject.  Currently gitsign doesn't offer any ability to specify annotations.
3. The spec wants the `subject.uri` to be set. Currently gitsign doesn't offer any ability to set the `subject.uri` field.
4. The spec wants the statement type to be set to `"https://in-toto.io/Statement/v1"` but currently gitsign uses `"https://in-toto.io/Statement/v0.1"`

It might be possible to resolve this by enable gitsign to take a 'raw' in-toto statement, but as @wlynch points out, that runs the risk of decoupling the commit referenced in the attestation from where gitsign is actually storing the data.

Instead, perhaps we can address this by updating gitsign to:

1. Set `gitCommit` in addition to the existing `sha256` type (or alternatively have some flag to set it to a more modern mode).
2. Update the statement type to `"https://in-toto.io/Statement/v1"`
3. Add a command line flag for setting the subject URI (e.g. `--subject_uri https://github.com/sigstore/gitsign/commit/9fc97de55fc1926fa2a76f06b02c8531dcf6828d`)
4. Add a command line flag for adding subject annotations (e.g. `--subject_annotation "source_branches=foo,bar"`)

In the end some might be able to sign a SLSA Source VSA with a command like:

`gitsign attest -f vsa-predicate.json --type https://slsa.dev/verification_summary/v1 --subject_annotation "source_branches=foo,bar" --subject_uri https://github.com/sigstore/gitsign/commit/9fc97de55fc1926fa2a76f06b02c8531dcf6828d`

Would something like this be reasonable?

**Version**
gitsign version v0.12.0

[...]

Issue #537, opened 2024-07-12
Using sigstore-go's API instead of Cosign


haydentherapper:
We, being the sigstore-go maintainers, have released a new version of sigstore-go with support for signing artifacts. We've been looking for good candidates in the ecosystem to try out sigstore-go, and I think gitsign would be a perfect choice. You're using Cosign's API currently for blob signing and not container signing, which means you should be able to migrate over to sigstore-go without any loss of functionality and a gain of fewer dependencies and a much cleaner API.

The main changes to note, beyond the API, are:

* The usage of a new TUF client that uses a [trust root bundle](https://github.com/sigstore/protobuf-specs/blob/main/protos/sigstore_trustroot.proto) to fetch roots of trust. This also simplifies support for private Sigstore instances, which can provide trust root material via a single file rather than needing a TUF environment or setting multiple environment variables
* Support for the new [bundle format](https://github.com/sigstore/protobuf-specs/blob/main/protos/sigstore_bundle.proto) for verification material (cert and proof) and signature.

Here is an example of how to use the API to sign - https://github.com/sigstore/sigstore-go/blob/main/examples/sigstore-go-signing/main.go - and to verify - https://github.com/sigstore/sigstore-go/blob/main/cmd/sigstore-go/main.go.

If you were to make this change, at a glance over the code, you would roughly need to:

* Fetch an identity token using Cosign's identity providers (we're talking about pulling these into a dedicated repo)
* Change [initialize](https://github.com/sigstore/gitsign/blob/359a77d4426e3d3196bd0f6da1520fa5fa20e376/internal/commands/initialize/initialize.go#L71) to initialize the new [TUF client](https://github.com/sigstore/sigstore-go/blob/main/pkg/tuf/client.go)
* Provide the identity token and artifact to sigstore-go's signing API, output a bundle, and transform the bundle into the PKCS7 structure you store in the commit
  * You can let the signing API handle [uploading to Rekor](https://github.com/sigstore/gitsign/blob/359a77d4426e3d3196bd0f6da1520fa5fa20e376/pkg/rekor/rekor.go#L109) and fetching a cert from Fulcio, the API just needs an identity token
[...]

PR #620, opened 2025-01-19
Modernize attest statements

TomHennen:
#### Summary

`attest` now generates 'modern' in-toto attestations using the [in-toto attestation go bindings](https://github.com/in-toto/attestation/tree/main/go) and dropped usage of the cosign library for statement generation.

This lets us:

1. Use [recommended `gitCommit`](https://github.com/in-toto/attestation/blob/main/spec/v1/digest_set.md#gitcommit-gittree-gitblob-gittag) as the digest type (which is what we were doing anyways)
2. Use the correct `_type` value for modern in-toto statements

In the future it will let us add annotations on the subject.

Note that this also changes the user behavior in that it:
1. Changes the things mentioned above.
2. Doesn't let users use the convenience attestation types from the command line.  They must instead specify the full predicate type.
    - It doesn't seem worthwhile to replicate cosign's helper method here, especially if we're being asked to stop using the cosign API anyways (#537).

refs #611

#### Release Note

* `attest` produces in-toto compliant statements

#### Documentation

No documentation updates needed as the `attest` command is not referenced in the current documentation. https://docs.sigstore.dev/cosign/signing/gitsign/

### DIFF
--- a/internal/attest/attest.go
+++ b/internal/attest/attest.go
@@ -47,8 +47,10 @@ import (
 )
 
 const (
-	CommitRef = "refs/attestations/commits"
-	TreeRef   = "refs/attestations/trees"
+	CommitRef        = "refs/attestations/commits"
+	TreeRef          = "refs/attestations/trees"
+	DigestTypeCommit = "gitCommit"
+	DigestTypeTree   = "gitTree"
 )
 
 var (
@@ -59,18 +61,20 @@ var (
 type rekorUpload func(ctx context.Context, rekorClient *rekorclient.Rekor, signature []byte, pemBytes []byte) (*models.LogEntryAnon, error)
 
 type Attestor struct {
-	repo    *git.Repository
-	sv      *sign.SignerVerifier
-	rekorFn rekorUpload
-	config  *gitsignconfig.Config
+	repo       *git.Repository
+	sv         *sign.SignerVerifier
+	rekorFn    rekorUpload
+	config     *gitsignconfig.Config
+	digestType string
 }
 
-func NewAttestor(repo *git.Repository, sv *sign.SignerVerifier, rekorFn rekorUpload, config *gitsignconfig.Config) *Attestor {
+func NewAttestor(repo *git.Repository, sv *sign.SignerVerifier, rekorFn rekorUpload, config *gitsignconfig.Config, digestType string) *Attestor {
 	return &Attestor{
-		repo:    repo,
-		sv:      sv,
-		rekorFn: rekorFn,
-		config:  config,
+		repo:       repo,
+		sv:         sv,
+		rekorFn:    rekorFn,
+		config:     config,
+		digestType: digestType,
 	}
 }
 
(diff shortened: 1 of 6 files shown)


TomHennen:
Hmm, these tests worked yesterday, perhaps there's some non-determinism in how the statement gets marshaled.  Let me check on that.

TomHennen:
Ok, I believe the tests have been fixed.  I'm not sure if folks would prefer this had been handled some other way.  Please let me know.

wlynch (approved):
Thanks!

internal/attest/attest.go
```diff
@@ -226,18 +228,32 @@ func encode(store storage.Storer, enc Encoder) (plumbing.Hash, error) {
 	return store.SetEncodedObject(obj)
 }
 
+func generateStatement(pred []byte, attType string, sha plumbing.Hash) (*spb.Statement, error) {
+	sub := []*spb.ResourceDescriptor{{
+		Digest: map[string]string{"gitCommit": sha.String()},
```
wlynch, line 233: This we're probably going to need to plumb through some additional data to determine if this should be gitCommit, gitTree, etc. but i'm okay with starting with this.

wlynch (dismissed):

internal/attest/attest.go
TomHennen: Right, that will also require changing how that templating works.  I'm somewhat wondering if it would be easier to drop the file based templating and instead construct the JSON 'inline' in code.  WDYT?

wlynch: This is for the test output? yeah that's fine with me!

TomHennen: Let me fix that now.

internal/commands/attest/attest.go
```diff
@@ -46,7 +46,7 @@ type options struct {
 func (o *options) AddFlags(cmd *cobra.Command) {
 	cmd.Flags().StringVar(&o.FlagObjectType, "objtype", FlagObjectTypeCommit, "[commit | tree] - Git object type to attest")
 	cmd.Flags().StringVarP(&o.FlagPath, "filepath", "f", "", "attestation filepath")
-	cmd.Flags().StringVar(&o.FlagAttestationType, "type", "", `specify a predicate type (slsaprovenance|link|spdx|spdxjson|cyclonedx|vuln|custom) or an URI (default "custom")`)
```
wlynch, line 49: After responding to https://github.com/sigstore/gitsign/issues/623#issuecomment-2608287770 this made me realize there was a good reason to have this, which was to support making it easy to add attestations for arbitrary files that don't fit neatly into an in-toto type, e.g. being able to do something like `gitsign attest -f test-results.xml --type custom` and have that work auto-magically.

That said, I think there's more benefits to getting this working with the new in-toto libraries at the moment so happy to proceed with this for now, but this will likely be something we'll look to add back in some form (either as this flag or another flag/command).

wlynch (approved):

PR #632, opened 2025-02-13
Fix test for go-git update

adityasaky:
#### Summary

This PR is a pass at fixing test failure from a recent go-git update.

#### Release Note

None

#### Documentation

None


jas4711:
Thank you!  I built this locally and can confirm it solves the problem with go-git 5.13.0.  We'll use this patch for gitsign in Debian.  Self-checks also works with go-git 5.12.0 so the patch should be safe to apply for anyone regardless of go-git version.

PR #657, opened 2025-03-25
Enable signing with --detached-sign flag.

wlynch:
#### Summary

[jj recently added gpgsm signing support](https://github.com/jj-vcs/jj/pull/6078), but @kevinmdavis pointed out to me that gitsign doesn't work out of the box. This is because [jj invokes gpg commands with `-abu`](https://github.com/jj-vcs/jj/blob/2271c306f0ea93dbfe2e62ffe299f300305c2c51/lib/src/gpg_signing.rs#L276-L282) instead of [git's `-bsau`](https://github.com/git/git/blob/683c54c999c301c2cd6f715c411407c413b1d84e/gpg-interface.c#L982-L985).

`-s` = `--sign`, but `-b` = `--detached-sign` which feels like should also enable signing mode.
This change allows signing to be done if either flag is set.

#### Release Note

Gitsign will sign if either `--sign` or `--detached-sign` is provided.

#### Documentation

n/a - should have no noticeable change. Might add jj docs later after more testing.

### DIFF
--- a/internal/commands/root/root.go
+++ b/internal/commands/root/root.go
@@ -76,7 +76,7 @@ func New(cfg *config.Config) *cobra.Command {
 							return item.RunE(item, cmd.Flags().Args())
 						}
 					}
-				case o.FlagSign:
+				case o.FlagSign, o.FlagDetachedSignature:
 					return commandSign(o, s, args...)
 				case o.FlagVerify:
 					return commandVerify(o, s, args...)


adityasaky (approved):
Interesting, the detached signature flag was triggering the help all this while? 🤔

wlynch:
> Interesting, the detached signature flag was triggering the help all this while? 🤔

Really the lack of `-s`, but I can see why a user would assume passing only `-b` would be sufficient.

Issue #736, opened 2026-01-13
Expose the attestation generator


puerco:
In order to generate gitsign attestations from other programs, we should formalize the attestation generator as a public package (right now it's implemented in private functions in the show CLI command).

PR #737, opened 2026-01-13
Expose predicate builder

puerco:
#### Summary

This PR exposes the attestation generator, moving the existing logic from the `show` subcommand to a new `attest` public package. I've made a couple of updates in addition to moving the code:

- The statement is now built with the newer [in-toto/attestation](https://github.com/in-toto/attestation) package.
- The predicate types are now built from protocol buffers definitions

Functionally, the statement generator is exactly the same except for two minor "breaking" changes:

1. The in-toto statement type URI is now updated from https://in-toto.io/Statement/v0.1 to the current https://in-toto.io/Statement/v1
2. The timestamps are now rendered in the zulu (Z) timezone.

Closes https://github.com/sigstore/gitsign/issues/736

#### Release Note
- Added new public pacakge exposing the attestaion generation logic.
- The attestation predicate type is now generated from protocol buffer definitions

#### Documentation
N/A

### DIFF
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -43,33 +43,14 @@ jobs:
         run: |
           set -e
 
-          # Detect platform and architecture
-          OS=$(uname -s | tr '[:upper:]' '[:lower:]')
-          ARCH=$(uname -m)
-          case $ARCH in
-            x86_64) ARCH="amd64" ;;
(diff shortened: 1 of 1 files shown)


puerco:
OK, I've pushed a small change to the tests from last night to build gitsign at the PR commit instead of pulling the last binary (which was emulating the previous behavior).

PR #738, opened 2026-01-13
Fix e2e ci

puerco:
#### Summary

This PR attempts to fix the CI failures by introducing three changes:

1. Adds a missing cd into the test git reposotory
2. Installs the compiled gitsign binary from the releases page to prevent conflicts when bumping go versions, especially building the previous gitsign version
3. Fixes a couple of linter problems
4. Frees some disk space in the runner by removing some crap before the goreleaser job that was [failing with no space left on device](https://github.com/sigstore/gitsign/actions/runs/20976688030) messages.

### DIFF
--- a/.github/workflows/validate-release.yml
+++ b/.github/workflows/validate-release.yml
@@ -27,6 +27,22 @@ jobs:
           restore-keys: |
             ${{ runner.os }}-go-
 
+      - name: Free disk space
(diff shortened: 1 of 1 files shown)


puerco:
OK, it seems those were all. Back to green now
/cc @wlynch

PR #802, opened 2026-05-06
Verify against raw git object bytes

wlynch:
#### Summary

go-git's loose object parser uses last-wins semantics for duplicate singleton headers (tree, author, committer, etc.), while git-core uses first-wins. An attacker can craft a commit or tag whose raw bytes hash to one set of contents under git-core but re-encode through go-git to a different signed payload, letting a legitimate signature verify against attacker-controlled bytes.

Replace the go-git decode + EncodeWithoutSignature path with SplitCommit and SplitTag, which operate directly on the object-database bytes (the same bytes git-core feeds its verifier) and reject objects with structural ambiguities — duplicate singleton headers, duplicate gpgsig, malformed gpgsig continuations. ObjectHash now reassembles via JoinCommit/JoinTag so the recorded hash matches git-core.

#### Release Note

- Fixes object parsing and validation for invalid Git objects (multiple/conflicting headers)

#### Documentation

n/a


cpanato (approved):

wlynch:
Merging w/o E2E because github actions outage has broken the Sigstore beacon token workflow, blocking the tests.

PR #819, opened 2026-06-02
Add experimental sigstore-go support

wlynch:
#### Summary

Adds a new config option - `gitsign.enableSigstoreGo` if set to `true`, then new sigstore-go signer/verifier is used to do the signing and verification of commits.

At a high level, this works by adding a compatibility layer between Sigstore Bundles and gitsign's CMS signature layout (which we use to stay compatible with other Git tooling). When signing, we get the bundle then convert it to a CMS signature. When verifying we construct the bundle from the CMS signature, then use sigstore-go to verify.

CAVEATS: For now, signing requires use of offline signatures. If verification encounters an online signature, we fall back to the old non-sigstore-go behavior. I'll probably add online signing/verification support in another PR.

Part of https://github.com/sigstore/gitsign/issues/537

#### Release Note

- Adds a new experimental config option - `gitsign.enableSigstoreGo` if set to `true`, then new sigstore-go signer/verifier is used to do the signing and verification of commits. This should not have a meaningful visible effect for most users.

#### Documentation

✅


internal/signature/sign.go
```diff
@@ -106,6 +123,17 @@ func Sign(ctx context.Context, ident Identity, body []byte, opts SignOptions) (*
 		}
 	}
 
+	// Experimental bundle signing path: sigstore-go produces the signature and
+	// Rekor entry as a bundle, which is then converted to CMS for storage. Only
+	// used in offline Rekor mode (opts.Rekor set).
+	if opts.Bundle && opts.Rekor != nil {
```
adityasaky, line 129: is opts.Rekor used? We use opts.RekorURL to build a rekor client but I'm not seeing signBundle use opts.Rekor either. Should this guard be for RekorURL instead?

docs/bundle-cms.md
adityasaky, line 1: If we have sigstore go enabled, will the new verification flow fail for commits created before sigstore go was enabled? Can we provide a fallthrough to users transparently for verification (and to clients like gittuf who use gitsign for verification)?

internal/signature/bundlesign.go
```diff
@@ -138,18 +137,16 @@ func (c *validatingRekorClient) CreateLogEntry(params *entries.CreateLogEntryPar
...
-	}
-	bb, err := x509.MarshalPKIXPublicKey(b)
-	if err != nil {
-		return err
+// samePublicKey reports whether two public keys are equal, using the key's own
+// Equal method (implemented by all stdlib public key types).
+func samePublicKey(a, b crypto.PublicKey) error {
+	key, ok := a.(interface {
+		Equal(x crypto.PublicKey) bool
+	})
```
adityasaky, line 145: TIL you can type assert inlined interfaces like so :O

docs/bundle-cms.md
adityasaky: nvm, I just got to ac217bde402828aaeddbd633244c84a07874f661

adityasaky:
@wlynch this largely looks fine to me. Only question is https://github.com/sigstore/gitsign/pull/819#discussion_r3364298683 which I might have misunderstood.

internal/signature/sign.go
wlynch: It's really a proxy for "are we doing offline signing".

It's confusing because we're in this in-between state. opts.Rekor is set when we do offline mode signing, but we can't use that because it masks the RekorURL so we need to use the other field. Eventually this will be all collapsed down to 1 client.

adityasaky (approved):

PR #821, opened 2026-06-03
Print signing identity and issuer to stderr after signing

### LINKED ISSUE #693: When signing, gitsign should show the certificate-oidc-issuer & certificate-identity used after signing complete
In version 2 of gitsign, the flags `--certificate-oidc-issuer` & `--certificate-identity` are now required to be passed to the `gitsign verify` command to tell gitsign which identity and issuer is expected to be valid. That's fine, and I understand why this decision was made. The problem is that it not always obvious to the user when signing a commit or tag what the URI for the `certificate-oidc-issuer` is, or which `certificate-identity` was used. For example, I signed a commit using my GitHub identity. When I went to verify, I was unsure what URI to pass for `--certificate-oidc-issuer`. I tried `https://github.com`, but that didn't work. It took some digging to figure out that what I wanted to pass was `https://github.com/login/oauth`. Also, I initially assumed the value I should pass for my `--certificate-identity` would be the GitHub privacy-sparing email. I didn't realize that it would use my private email, which meant I was passing the wrong value to `--certificate-identity` as well.

I suggest that, after a user signs a commit or tag, gitsign displays the value for `certificate-oidc-issuer` and `certificate-identity` that was just used to sign. That will make it obvious to the user which values to pass to the flags when verifying. I also recommend adding a command to `gitsign` to allow a user to inspect the value for `--certificate-identity` & `--certificate-oidc-issuer` that was used for a commit without actually verifying it. I understand that you want to ensure that people don't just blindly trust a verification without knowing *who* actually did the signing, however in an open-source project with multiple contributors it can be cumbersome to try to guess the correct identity for any particular commit, particularly since people who commit using their GitHub privacy-sparing e-mail addresses will show that address in the regular Git commit log, but if someone wants to verify using gitsign they would need to pass in that user's private e-mail address, which he/she may not know.

arpitjain099:
## Summary

After a successful sign, gitsign now prints the `certificate-identity` and `certificate-oidc-issuer` that were actually used, plus a ready-to-run `gitsign verify` command. This makes it obvious which values to pass to the verify flags.

This is the narrower half of #693 (report-after-signing). It does not add a separate "inspect a commit" subcommand.

Fixes #693

## Why

In v2, `gitsign verify` requires `--certificate-identity` and `--certificate-oidc-issuer`, but it is not obvious which values to use. As the issue notes, the GitHub issuer is `https://github.com/login/oauth` (not `https://github.com`), and the identity is the user's private email rather than the privacy-sparing address. Surfacing both right after signing removes that guesswork.

## Output goes to stderr, not stdout

@adityasaky raised the concern about tools that parse gitsign's stdout after signing. To avoid breaking those, the new message is written to `TTYOut`, which is a TTY when one is available and falls back to stderr otherwise, and never to `s.Out`. `s.Out` (stdout) still carries only the signature bytes, so machine parsers consuming gitsign's stdout are unaffected. This is the same stream the existing "tlog entry created with index" status line already uses.

Example (stderr):

```
gitsign: signed with identity "foo@example.com" (issuer "https://github.com/login/oauth")
gitsign: to verify, run: gitsign verify --certificate-identity "foo@example.com" --certificate-oidc-issuer "https://github.com/login/oauth"
```

## How identity and issuer are extracted

A small helper, `internal.NewSigningIdentity(cert)`, reads:

- the identity from the certificate subject alternative names via `cryptoutils.GetSubjectAlternateNames` (joining multiple SANs so nothing is silently dropped), and
- the issuer from the Fulcio OIDC issuer extension via `cosign.CertExtensions{Cert: cert}.GetIssuer()`.

These are the same sources `gitsign verify` already reads back when it prints "Good signature from ...", so what is reported at sign time matches what verify expects.

## Tests

`internal/utils_test.go` adds table-driven unit tests for the extraction (email SAN, URI SAN, multiple SANs joined, and the empty cert) and for the formatted guidance string. `go test ./internal/...` passes; `go vet` and `gofmt` are clean.

### DIFF
--- a/internal/utils.go
+++ b/internal/utils.go
@@ -19,7 +19,12 @@ import (
 	"crypto/sha1" // #nosec G505
 	"crypto/x509"
 	"encoding/hex"
+	"fmt"
 	"net/url"
+	"strings"
+
+	"github.com/sigstore/cosign/v3/pkg/cosign"
+	"github.com/sigstore/sigstore/pkg/cryptoutils"
 )
 
 // certHexFingerprint calculated the hex SHA1 fingerprint of a certificate.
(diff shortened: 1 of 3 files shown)


internal/utils.go
```diff
@@ -37,6 +42,42 @@ func certFingerprint(cert *x509.Certificate) []byte {
...
+	return SigningIdentity{
+		Identity: identity,
+		Issuer:   ce.GetIssuer(),
+	}
+}
+
+// String formats the signing identity as guidance for the verify command,
+// pointing the user at the exact flag values to pass.
+func (s SigningIdentity) String() string {
+	return fmt.Sprintf("gitsign: signed with identity %q (issuer %q)\ngitsign: to verify, run: gitsign verify --certificate-identity %q --certificate-oidc-issuer %q",
```
wlynch: First line looks good, but the second feels somewhat verbose. I'd drop this or put it behind a flag so it's not shown by default.

wlynch (commented):
Thanks for the commit! One small comment.

arpitjain099:
Dropped the second line (the full `gitsign verify ...` command) as suggested - it now prints just the identity and issuer. Thanks @wlynch!

internal/utils.go
arpitjain099: Bumping this gently since it's been a while. I dropped the second line back in June (4755721), so the output is now just the identity and issuer, which I think is what you were after.

No rush at all, but let me know if you'd like anything else changed here. Happy to rebase onto main too if that helps, the branch is a few commits behind now.

arpitjain099: Rebased onto main just now (98340df), so it's up to date and conflict-free. `go build ./...` and `go test ./internal/...` are both green on the rebased branch, including with the sigstore-go 1.2.0 bump that landed in the meantime.

Ready whenever you or another maintainer has a chance to take a look.

PR #825, opened 2026-06-09
fulcio: add 'device' TokenProvider for OAuth 2.0 device-code flow

dustinkirkland:
## Summary

Adds `device` as a new value for `GITSIGN_TOKEN_PROVIDER` / `gitsign.tokenProvider`. When set, gitsign performs an RFC 8628 [device authorization grant](https://datatracker.ietf.org/doc/html/rfc8628) directly against `cfg.Issuer` (default `https://oauth2.sigstore.dev/auth`) using the sigstore library's existing [`oauthflow.NewDeviceFlowTokenGetterForIssuer`](https://github.com/sigstore/sigstore/blob/main/pkg/oauthflow/device.go).

The user opens the printed verification URL on any browser — phone, laptop, second machine, anywhere — to complete OAuth consent. The SSH session itself never needs a browser, port forward, or localhost callback. This unblocks signing on remote / headless developer hosts (long-running SSH sessions, dev VMs with no GUI, etc.) where the existing interactive flow can't open a browser locally and the OOB fallback is clunky.

## Implementation

Parallel to the existing `interactive` special case in `internal/fulcio/identity.go`: `device` sets `authFlow` directly to a `DeviceFlowTokenGetter` rather than routing through the cosign providers registry and the `StaticTokenGetter` shim. The other registry-backed providers (`spiffe`, `google-impersonation`, `filesystem`, etc.) are unchanged.

Three files changed, +17 / -3 lines total:
- `internal/fulcio/identity.go` — the new `else if cfg.TokenProvider == \"device\"` branch
- `internal/config/config.go` — comment updated to list `device` as a valid value with a one-paragraph description
- `README.md` — `tokenProvider` config row and `GITSIGN_TOKEN_PROVIDER` env-var row both updated

## How I validated

Built locally and verified the test suite still passes:

\`\`\`
$ go build ./...
$ env -u GITSIGN_OIDC_REDIRECT_URL -u GITSIGN_CONNECTOR_ID -u GITSIGN_TOKEN_PROVIDER \\
    go test ./internal/fulcio/... ./internal/config/...
ok  	github.com/sigstore/gitsign/internal/fulcio/fulcioroots	(cached)
ok  	github.com/sigstore/gitsign/internal/config	0.003s
\`\`\`

End-to-end: I'm running a small CLI tool that uses the same `oauthflow.NewDeviceFlowTokenGetterForIssuer` API to mint an OIDC token and write it to disk, then signing commits via the existing filesystem token provider. With this PR that helper becomes redundant — `GITSIGN_TOKEN_PROVIDER=device git commit` does the whole thing in-process.

## Related

I have a companion PR at https://github.com/sigstore/cosign/pull/4934 that lets the cosign filesystem provider honor a `SIGSTORE_OIDC_TOKEN_FILE` env var. Bundled framing for both: *\"headless desktop sigstore signing without sudo\"*. The two PRs are independent and either can land first.

A possible follow-up (not in this PR): the underlying `DeviceFlowTokenGetter` doesn't currently forward `connector_id` in the device-code request, so users land on Dex's provider-picker page before being redirected to the chosen IdP. The interactive flow already pre-selects via `ExtraAuthURLParams`; adding the same to the device flow would be a small sigstore-library PR.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


wlynch (approved):

wlynch:
Needs another approval since I can't self-approve the lint fixes.

cpanato (approved):

PR #828, opened 2026-06-16
Restore keyless attest build against cosign 3.1

### LINKED ISSUE #826: gitsign fails to build with cosign 3.1.x
Hi. I'm upgrading cosign in Debian and ran into a problem for gitsign, it fails like this:

```
# github.com/sigstore/gitsign/internal/commands/attest
src/github.com/sigstore/gitsign/internal/commands/attest/attest.go:78:36: undefined: signcommon.GetSignerVerifier
```

See full build log here: https://salsa.debian.org/jas/cosign/-/jobs/9753241#L5017

I thought this was a cosign API bug but they don't consider so: https://github.com/sigstore/cosign/issues/4923

Could you make gitsign work with latest cosign?

Thanks!

arpitjain099:
cosign 3.1 made `signcommon.GetSignerVerifier` private (part of the sign-via-sigstore-go rework in sigstore/cosign#4618), so the attest command stopped building - that is the #826 / Debian breakage @jas4711 hit.

The struct it returned (`signcommon.SignerVerifier`) and `fulcio.NewSigner` are both still exported, so I rebuilt what `GetSignerVerifier` did for the keyless case right in `attest.go`: generate an ephemeral key, get a Fulcio cert for it, and wrap both in a `SignerVerifier`. The attestor and its tests are untouched. Bumps cosign to v3.1.1.

`go build ./...` (after `go mod tidy`) is clean and the `internal/attest` tests pass. I couldn't exercise the real keyless Fulcio+Rekor path locally since it needs an interactive OIDC login, but the construction mirrors what cosign's own `GetSignerVerifier` did.

This is the minimal change to get the build green again. If you'd rather move the attestor onto cosign's new `GetKeypairAndToken` / sigstore-go keypair flow instead (the more future-proof route that came up in the issue), I'm happy to do that as a follow-up.

Fixes #826


arpitjain099:
Closing this as superseded.

This branch worked around cosign 3.1 unexporting `signcommon.GetSignerVerifier` by rebuilding the equivalent locally: ephemeral key, Fulcio cert, wrapped in a `signcommon.SignerVerifier`. Upstream has since taken a better route in #844 and the surrounding work, dropping the cosign signing path from `attest` altogether in favour of gitsign's own `internal/fulcio.NewIdentity` and `internal/rekor/tlog`. `signcommon` is no longer referenced anywhere in the tree.

That also means attestations now honour the same OIDC config and credential cache as commit signing, which my version did not.

I confirmed current `main` (6f3d985, cosign v3.1.2) builds clean with `go build ./...` and `go vet ./internal/commands/attest/...`, so there is nothing left here to carry forward.

PR #867, opened 2026-08-09
fix: wrap RFC3161 timestamp tokens as TimeStampResp in sigstore bundles

### LINKED ISSUE #866: Offline TSA timestamps unverifiable: bundle conversion writes bare TimeStampToken where protobuf-specs requires TimeStampResponse
### Description

In gitsign v0.17.0, offline signing with `GITSIGN_TIMESTAMP_SERVER_URL` set produces a CMS `id-aa-timeStampToken` unsigned attribute whose bytes are a bare RFC 3161 `TimeStampToken` (a CMS `ContentInfo` per RFC 5652 §11.4). At verify time, the bundle-conversion layer in `internal/sigstore/compat` reads those same bytes verbatim into the assembled bundle's `RFC3161SignedTimestamp.signed_timestamp` protobuf field. But the sigstore protobuf-specs spec that field as *"the DER encoded TimeStampResponse"* (a different ASN.1 structure — `SEQUENCE { PKIStatusInfo, TimeStampToken OPTIONAL }`, RFC 3161 §2.4.2).

Downstream, sigstore-go's TSA verifier tries to parse the bytes as `TimeStampResp` and fails. Every offline-mode gitsign commit that includes a TSA timestamp is unverifiable through the standard verify path.

### Reproducer

Sign a commit against a real `sigstore/timestamp-authority` in offline + sigstore-go mode:

```bash
export GITSIGN_ENABLE_SIGSTORE_GO=true
export GITSIGN_REKOR_MODE=offline
export GITSIGN_REKOR_VERSION=2
export GITSIGN_REKOR_URL=http://your-rekor-tiles/
export GITSIGN_FULCIO_URL=http://your-fulcio/
export GITSIGN_TIMESTAMP_SERVER_URL=http://your-tsa/api/v1/timestamp
git commit -S -m "..."
```

Then verify offline against the standard trust files:

```bash
SIGSTORE_ROOT_FILE=fulcio.pem \
SIGSTORE_REKOR_PUBLIC_KEY=rekor-pub.pem \
SIGSTORE_CT_LOG_PUBLIC_KEY_FILE=ctlog-pub.pem \
GITSIGN_TIMESTAMP_CERT_CHAIN=tsa-root.pem \
gitsign verify --certificate-identity=<expected> --certificate-oidc-issuer=<expected> HEAD
```

Result:

```
Error: no signer could be verified: signer 0: failed to verify timestamps: threshold not met for verified signed timestamps: 0 < 1; error: unable to verify signed timestamps: error parsing response into Timestamp: asn1: structure error: tags don't match (16 vs {class:0 tag:6 length:9 isCompound:false}) {optional:false explicit:false application:false private:false defaultValue:<nil> tag:<nil> stringType:0 timeType:0 set:false omitEmpty:false} pkiStatusInfo @2
```

The `pkiStatusInfo @2` error is the giveaway: `TimeStampResp` expects a `SEQUENCE` at field position 2 (the PKIStatusInfo), but a bare `TimeStampToken` has an `OBJECT IDENTIFIER` there (contentType of the ContentInfo).

### Source trace

Sign side (correct per RFC 5652 §11.4 — a bare `TimeStampToken` in `id-aa-timeStampToken`):
- `internal/fork/ietf-cms/timestamp.go:14-37` — `SignedData.AddTimestamps(url)` fetches TSA response, extracts `resp.TimeStampToken`, stores it in `AttributeTimeStampToken`.
- `internal/signature/sign.go:161-163` and `internal/signature/bundlesign.go:85-87` — signing paths that call `AddTimestamps`.

Verify-side conversion (where the shape mismatch surfaces):
- `internal/sigstore/compat/attrs.go:29-44` — `timestampTokens(si)` reads `si.UnsignedAttrs.GetValues(oid.AttributeTimeStampToken)` and returns the value bytes verbatim.
- `internal/sigstore/compat/compat.go:146-158` — those bytes are appended to `bundle.VerificationMaterial.TimestampVerificationData.Rfc3161Timestamps[*].SignedTimestamp` without any shape conversion.

Spec (`sigstore/protobuf-specs`) — `protos/sigstore_common.proto`:
```
message RFC3161SignedTimestamp {
        // Signed timestamp is the DER encoded TimeStampResponse.
        // See https://www.rfc-editor.org/rfc/rfc3161.html#section-2.4.2
        bytes signed_timestamp = 1 [(google.api.field_behavior) = REQUIRED];
}
```

Downstream verify (correct per spec, so it rejects the bare token):
- `sigstore-go/pkg/root/timestamping_authority.go` — `SigstoreTimestampingAuthority.Verify` calls `tsaverification.VerifyTimestampResponse`.
- `sigstore/timestamp-authority/pkg/verification/verify.go:VerifyTimestampResponse` — calls `github.com/digitorus/timestamp.ParseResponse`.
- `digitorus/timestamp/timestamp.go:ParseResponse` — `asn1.Unmarshal(bytes, &resp)` against the `response` struct (SEQUENCE of `PKIStatusInfo` + optional `TimeStampToken`). A bare `TimeStampToken` starts with an OID where PKIStatusInfo expects a SEQUENCE, so this fails at field position 2.

### Fix directions

Either:

1. **Wrap at bundle-conversion time** — in `internal/sigstore/compat/attrs.go:timestampTokens`, wrap each raw attribute value in a synthetic `TimeStampResp` (`SEQUENCE { SEQUENCE { INTEGER 0 (granted) }, TimeStampToken }`) before returning. The added overhead is 5-9 bytes per attribute. Preserves the existing CMS sign-side behavior (which is RFC-compliant), fixes only the bundle field to match its protobuf-specs contract.

2. **Capture the full response at sign time** — in `internal/fork/ietf-cms/timestamp.go:fetchTS` and `internal/signature/bundlesign.go`, embed the whole TSA HTTP response body (which IS a `TimeStampResp`) into the CMS attribute rather than extracting and storing just the token. This makes the CMS attribute non-compliant with RFC 5652 §11.4 (id-aa-timeStampToken is defined as a bare `TimeStampToken`), which may break other consumers.

Option 1 seems cleaner and less invasive.

### Environment

- gitsign v0.17.0 (bug also present in unreleased HEAD as of 2026-08-06)
- sigstore-go v1.2.2
- cosign v3.1.2
- sigstore/timestamp-authority v2.1.3 (live TSA)
- Rekor v2 (rekor-tiles v2.3.0)
- Fulcio v1.8.8

No workaround available on the verify side — `--insecure-ignore-sct` doesn't affect timestamp thresholds, and `WithObserverTimestamps(1)` is unconditionally added to the verifier options in `internal/gitsign/bundle.go:369-380` (no CLI or env knob to disable).

### Note

We implemented fix direction 1 externally as a wrapper around
gitsign — wrapping the raw attribute value in a synthetic
`TimeStampResp` before the commit is signed. The resulting
commits verify with stock `gitsign verify`, offline, with no
network access. Happy to open a PR if the maintainers prefer
that shape.

pujitha24:
Motivation:
When gitsign signs with a timestamp authority and sigstore-go bundle
conversion, the CMS unsigned attribute id-aa-timeStampToken correctly
stores a bare RFC 5652 SS11.4 TimeStampToken (a CMS ContentInfo). But
internal/sigstore/compat's bundle conversion copied those bytes
verbatim into the bundle's RFC3161SignedTimestamp.SignedTimestamp
field, which sigstore's protobuf-specs defines as a DER-encoded
TimeStampResp (SEQUENCE{PKIStatusInfo, TimeStampToken OPTIONAL}) -
a different ASN.1 structure. Downstream verifiers (sigstore-go,
digitorus/timestamp) fail to parse the bare token as a TimeStampResp
with an ASN.1 tag mismatch on the PKIStatusInfo field, so any bundle
carrying a timestamp fails threshold verification.

Approach:
timestampTokens (internal/sigstore/compat/attrs.go) now parses each
raw attribute value as a protocol.ContentInfo and re-encodes it inside
a timestamp.Response{Status: granted, TimeStampToken: <token>} before
returning it. timestamp.Response is this repo's existing, already
asn1-tagged TimeStampResp type (internal/fork/ietf-cms/timestamp),
whose round-trip DER shape is already covered by an existing test.
Hardcoding a granted status is not a fabrication: the sign path
(internal/fork/ietf-cms/timestamp.go's fetchTS) only stores the
attribute after the original TSA response's status was already
checked as granted, and the actual token content (signature, cert,
genTime) is preserved byte-for-byte, so downstream signature and
cert-chain verification is unaffected.

Validation:
Added internal/sigstore/compat/attrs_test.go: TestTimestampTokens
builds a synthetic CMS unsigned attribute holding a bare ContentInfo,
calls timestampTokens, and asserts the returned bytes parse as a
timestamp.Response with the expected fields. Confirmed this test
fails against the pre-fix code with the same ASN.1 tag-mismatch error
class reported in the issue (a structure error on the PKIStatusInfo
field) and passes after the fix. Ran:

  go build ./...
  go test ./internal/sigstore/compat/... ./internal/fork/ietf-cms/... -v
  go test ./...

All pass. This proves the ASN.1 shape defect and its fix directly; it
does not reproduce the full live-TSA end-to-end flow from the report
(no network access here), but the digitorus/timestamp parser used
downstream is a plain DER decoder with no network-observable state,
so the unit-level fix is sufficient to resolve the parse failure
described in the report.

Report: https://github.com/sigstore/gitsign/issues/866
Signed-off-by: Pujitha Paladugu <10557236+pujitha24@users.noreply.github.com>

Fixes #866


Hammu-admin:
I filed #866. I've built this PR and run it end-to-end against a fully self-hosted Sigstore stack — Fulcio, Rekor v2 (tile-based), a tile-based CT log, and a self-hosted RFC 3161 timestamp authority, on ARM64, verified offline. That's the path noted as untestable here, so here are the results.

Built from 0b7989d for linux/arm64. Verification runs inside a container with --network=none, preceded by an isolation check against three targets, using only four public trust files.

signer → verifier	           stock v0.17.0	                                patched (this PR)
stock v0.17.0	                   FAIL — pkiStatusInfo @2	                 PASS, exit 0
patched	                            FAIL — pkiStatusInfo @2	                  PASS, exit 0

The fix resolves the issue end-to-end. Validated Git signature: true / Validated Rekor entry: true / Validated Certificate claims: true, exit 0.

The stock-signs / patched-verifies cell is worth calling out. It passes, which means commits already in history — signed by unpatched gitsign with a TSA in the path — become verifiable once this merges. No re-signing required. I confirmed via openssl asn1parse that the CMS bytes are unchanged between the two signers: both carry a bare ContentInfo under id-smime-aa-timeStampToken. The fix is entirely in the CMS→bundle conversion, which is what makes it retroactive.

One correction to my own report. The Note at the bottom of #866 says our external workaround implements fix direction 1. That's wrong, and I should have caught it before filing. Our wrapper operates on the sign side — it rewrites the unsigned attribute so the commit object itself carries TimeStampResp shape. That's closer to the direction 2 I described, and it's why our commits verify against stock gitsign of any version.

The distinction matters for what this PR does and doesn't cover. Direction 1 keeps the stored CMS attribute compliant with RFC 5652 §11.4, which is correct, and fixes gitsign's own verifier. But a third-party consumer that reads the CMS attribute directly and assembles its own bundle will still hit the same TimeStampResp mismatch, because the bytes on disk are unchanged. If universal verifiability by non-gitsign tooling is a goal, a sign-side companion change would be needed — wrapping resp.TimeStampToken before NewAttribute in internal/fork/ietf-cms/timestamp.go's fetchTS. That has the RFC 5652 §11.4 cost I flagged in the issue, so it's a real trade rather than an obvious improvement.

Either way, this fixes what #866 reported. Happy to run further cases against the live stack if useful.

pujitha24:
Thanks for testing this against a real stack — I don't have network access to a live TSA/Fulcio/Rekor setup here, so I can't reproduce those numbers myself, but good to hear it resolves the parse failure end-to-end. Appreciate the correction on the workaround direction too.

On the sign-side companion change: agreed that's a separate, real trade-off (giving up RFC 5652 §11.4 compliance on the stored CMS attribute) rather than a clear improvement, so I'll leave it out of this PR. This PR is scoped to the CMS→bundle conversion that #866 reported. If maintainers want the sign-side change for third-party consumers reading the raw attribute, that'd be worth its own issue/PR and discussion.

pujitha24:
This has been rebased and green for a week, with the underlying fix validated end-to-end against a live stack per the comment above — let me know if there's anything I can do to make review easier.

wlynch (approved):
Thanks!

